PR #4, opened 2022-07-19
[_]: Include the price currency when getting all the prices

PixoDev:
Include the price currency, so we don't hardcode this values in the client side


fabioespinosa (approved):
Seems ok to me

PR #6, opened 2022-08-09
[_]: Feat/send notification on subscription changed

fabioespinosa:
Sends a `post` request to notification service once a subscription has changed.

**Note: Before merge these env variables must be provided:**
- `NOTIFICATIONS_URL`
- `NOTIFICATIONS_API_KEY`


src/services/NotificationService.ts
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      const res = await axios.post(process.env.NOTIFICATIONS_URL as string, data, {
+        headers: { 'X-API-KEY': process.env.NOTIFICATIONS_API_KEY as string },
+      });
+      if (res.status !== 201)
+        console.warn(
+          `Post to notifications service failed with status ${res.status}. Data: ${JSON.stringify(data, null, 2)}`,
+        );
+    } catch (err) {
+      console.warn(
```
fabioespinosa, line 46: couldn't find a better way to use fastify.log here. Perhaps we could export the Fastify instance as a singleton from `app.ts`

Defaulted to console.

src/webhooks/index.ts
```diff
@@ -45,6 +45,7 @@ export default function (
           await handleSubscriptionCanceled(
```
fabioespinosa, line 45: Perhaps this switch has now grown big enough to have an event bus who handles these events (just a suggestion)

sg-gs: I would say a map serialized or an event bus could be useful, but I will not add that complexity increase rn

src/services/NotificationService.ts
sg-gs: we will revisit this in the future

sg-gs (approved):

sg-gs:
The environment variables ARE NOT still set. Set them before deploying @fabioespinosa or I can do it if required.

PR #11, opened 2022-10-10
[_]:(feature) Adding coupon code param into /checkout-session path

masterprog-cmd:
Coupon code added into /checkout-session path for obtain the discount

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -115,7 +115,7 @@ export default function (
     fastify.get('/subscriptions', async (req, rep) => {
       let response: UserSubscription;
 
-      const user = await assertUser(req, rep);
+      const user: User = await assertUser(req, rep);
 
       let subscriptionInCache: UserSubscription | null | undefined;
       try {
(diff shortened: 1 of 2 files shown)


src/controller.ts
```diff
@@ -178,7 +180,8 @@ export default function (
           req.body.price_id,
           req.body.success_url,
           req.body.cancel_url,
-          user ?? req.body.customer_email,
+          user?.customerId ?? req.body.customer_email,
```
fabioespinosa, line 183: estas seguro customerId lo podemos poner como customer_email?

fabioespinosa (approved):

src/services/PaymentService.ts
```diff
@@ -173,16 +173,18 @@ export class PaymentService {
...
       success_url: successUrl,
       cancel_url: cancelUrl,
       customer: typeof prefill === 'string' ? undefined : prefill?.customerId,
       customer_email: typeof prefill === 'string' ? prefill : undefined,
-      mode: 'subscription',
       line_items: [{ price: priceId, quantity: 1 }],
-      allow_promotion_codes: true,
+      mode: 'subscription',
+      discounts: couponCode ? [{ coupon: couponCode }] : undefined,
+      allow_promotion_codes: couponCode ? undefined : true,
```
fabioespinosa, line 187: This should be true anyway

PR #16, opened 2023-02-23
[_]: feat/add-optional-free-trial-param

masterprog-cmd:
A trial parameter has been added to allow the user to have free days/months.

### DIFF
--- a/src/services/PaymentService.ts
+++ b/src/services/PaymentService.ts
@@ -189,7 +189,6 @@ export class PaymentService {
       allow_promotion_codes: couponCode ? undefined : true,
       billing_address_collection: 'required',
       subscription_data: {
-        trial_from_plan: trial ? true : undefined,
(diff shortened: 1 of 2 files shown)


src/services/PaymentService.ts
```diff
@@ -188,6 +188,9 @@ export class PaymentService {
       discounts: couponCode ? [{ coupon: couponCode }] : undefined,
       allow_promotion_codes: couponCode ? undefined : true,
       billing_address_collection: 'required',
+      subscription_data: {
+        trial_period_days: trial ? 30 : undefined,
```
sg-gs, line 192: Add a trial or do not add it, but do not add a default trial, that would add trials to all subscriptions

sg-gs (changes requested):

sg-gs (approved):

PR #19, opened 2023-04-11
[PB-242]: Feat/provide comeback coupon

masterprog-cmd:
If the user didn't used the requested coupon, the function provides the coupon id to use the coupon


src/controller.ts
```diff
@@ -153,6 +153,10 @@ export default function (
       return paymentService.getPrices();
     });
 
+    fastify.get<{ Body: { email: string; coupon_code: string } }>('/apply-coupon', async (req, rep) => {
+      return paymentService.hasUserAppliedCoupon(req.body.email, req.body.coupon_code);
```
sg-gs, line 157: Please use camelCase. I know we have some snake case going on, but that is technical debt, we should use camel Case

sg-gs (changes requested):

sg-gs (approved):

PR #20, opened 2023-04-12
[PB-243]: feat/apply-coupon-to-given-user

masterprog-cmd:
Apply couponCode when the user is elegible and has an active subscription.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -164,12 +164,16 @@ export default function (
       const { uuid } = req.user.payload;
       const user = await usersService.findUserByUuid(uuid);
 
-      return paymentService.applyCouponToUser(user.customerId).catch((err) => {
+      try {
+        const applyCoupon = await paymentService.applyCouponToUser(user.customerId);
+        return applyCoupon;
+      } catch (err) {
         if (err instanceof CouponAlreadyAppliedError) {
           return rep.status(403).send({ message: 'Coupon already applied' });
+        } else {
+          return rep.status(500).send({ message: 'Internal server error' });
         }
-        throw err;
-      });
+      }
     });
 
     fastify.post<{
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -169,8 +169,9 @@ export default function (
         return applyCoupon;
       } catch (err) {
         if (err instanceof CouponAlreadyAppliedError) {
-          return rep.status(403).send({ message: 'Coupon already applied' });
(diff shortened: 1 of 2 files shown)


src/controller.ts
```diff
@@ -160,6 +160,22 @@ export default function (
...
+      const user = await usersService.findUserByUuid(uuid);
+
+      try {
+        const applyCoupon = await paymentService.applyCouponToUser(user.customerId);
+        return applyCoupon;
+      } catch (err) {
+        if (err instanceof CouponAlreadyAppliedError) {
+          return rep.status(403).send({ message: 'Coupon already applied' });
+        } else {
+          return rep.status(500).send({ message: 'Internal server error' });
```
sg-gs, line 174: Please log the error in case an Internal Error happens, so we can see what is happening.

src/services/PaymentService.ts
```diff
@@ -236,3 +254,4 @@ export class PaymentService {
 }
 
 class NotFoundSubscriptionError extends Error {}
+export class CouponAlreadyAppliedError extends Error {}
```
sg-gs, line 257: Please add the error message here: 
```
export class CouponAlreadyAppliedError extends Error {
  constructor() {
     super('User already applied coupon');
     
     Object.setPrototypeOf(this, CouponAlreadyAppliedError.prototype);
  }
}
```

The line 
```
Object.setPrototypeOf(this, CouponAlreadyAppliedError.prototype);
```

Is due to a TypeScript error: https://dannyguo.medium.com/how-to-fix-instanceof-not-working-for-custom-errors-in-typescript-1df978100a27

If you do not put it, `instanceof` will not work

src/controller.ts
```diff
@@ -160,6 +160,22 @@ export default function (
...
+    fastify.put('/apply-coupon', async (req, rep) => {
+      const { uuid } = req.user.payload;
+      const user = await usersService.findUserByUuid(uuid);
+
+      try {
+        const applyCoupon = await paymentService.applyCouponToUser(user.customerId);
+        return applyCoupon;
+      } catch (err) {
+        if (err instanceof CouponAlreadyAppliedError) {
+          return rep.status(403).send({ message: 'Coupon already applied' });
```
sg-gs, line 172: Better use what you already have: 
```
return rep.status(403).send({ message: err.message });
```

sg-gs (changes requested):

PR #21, opened 2023-04-19
[_]: fix/Fix amountAfterCoupon param

masterprog-cmd:
Fix the amoutAfterCoupon parameter to get the actual price of the next invoice.

### DIFF
--- a/src/services/PaymentService.ts
+++ b/src/services/PaymentService.ts
@@ -174,7 +174,7 @@ export class PaymentService {
       }
     }
 
-    const upcomingInvoice = await this.provider.invoices.retrieveUpcoming({ subscription: subscription.id });
+    const upcomingInvoice = await this.provider.invoices.retrieveUpcoming({ customer: customerId });
 
     const { price } = subscription.items.data[0];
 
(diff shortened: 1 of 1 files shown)


src/services/PaymentService.ts
```diff
@@ -184,7 +184,7 @@ export class PaymentService {
       currency: price.currency,
       interval: price.recurring!.interval as 'year' | 'month',
       nextPayment: subscription.current_period_end,
-      amountAfterCoupon: upcomingInvoice.total !== price.unit_amount ? upcomingInvoice.total : undefined,
+      amountAfterCoupon: upcomingInvoice.total,
```
masterprog-cmd, line 187: Maybe we can remove this parameter and add the upcomingInvoice.total in the "amount" parameter?

sg-gs: Lets keep this parameter as maybe in the future, we have the possibility to show before/after price open.

src/services/PaymentService.ts
```diff
@@ -174,7 +174,7 @@ export class PaymentService {
       }
     }
 
-    const upcomingInvoice = await this.provider.invoices.retrieveUpcoming({ subscription: subscription.id });
+    const upcomingInvoice = await this.provider.invoices.retrieveUpcoming({ customer: customerId });
```
sg-gs, line 177: Does changing the query affect any case?

sg-gs (approved):

PR #26, opened 2023-06-07
[_]: feat/check-unique-codes

masterprog-cmd:
New call added to check if the unique code is available or not to avoid creating user if it is not available.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -14,6 +14,16 @@ import {
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const rateLimit = require('fastify-rate-limit');
 
+//Type for allowed methods for each route
+type AllowedMethods = 'GET' | 'POST';
+
+const allowedRoutes: {
+  [key: string]: AllowedMethods[];
+} = {
+  '/prices': ['GET'],
+  '/is-unique-code-available': ['GET'],
+};
+
 export default function (
   paymentService: PaymentService,
   usersService: UsersService,
@@ -42,11 +52,8 @@ export default function (
     });
     fastify.addHook('onRequest', async (request, reply) => {
       try {
-        const config: { url?: string; method?: string } = request.context.config;
-        if (
-          (config.url && config.url === '/prices') ||
-          (config.url === '/is-unique-code-available' && config.method && config.method === 'GET')
-        ) {
+        const config: { url?: string; method?: AllowedMethods } = request.context.config;
+        if (config.method && config.url && allowedRoutes[config.url].includes(config.method)) {
           return;
         }
         await request.jwtVerify();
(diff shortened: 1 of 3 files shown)


src/controller.ts
```diff
@@ -6,14 +6,24 @@ import fastifyJwt from '@fastify/jwt';
...
-  LicenseCodesService 
+import {
+  InvalidLicenseCodeError,
+  LicenseCodeAlreadyAppliedError,
+  LicenseCodesService,
 } from './services/LicenseCodesService';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const rateLimit = require('fastify-rate-limit');
 
+//Type for allowed methods for each route
```
sg-gs, line 17: The type name explains itself, this comment is not necessary

src/controller.ts
```diff
@@ -252,19 +262,43 @@ export default function (
...
+        config: {
+          rateLimit: {
+            max: 5,
+            timeWindow: '1 minute',
+          },
+        },
+      },
+      async (req, res) => {
+        const { code, provider } = req.query;
+        await licenseCodesService.isLicenseCodeAvailable(code, provider);
```
sg-gs, line 283: This endpoint has a failure that could be annoying for the server. This use case throws if the code is redeemed or not found. I suggest catching here those errors and returning the correspondent 404 code. If not, a 500 is going to be thrown, which is not semantic and false, as there is no internal server error happening rather than a 404 that we are not handling properly.

sg-gs (changes requested):

src/controller.ts
masterprog-cmd: Done

src/controller.ts
```diff
@@ -252,19 +261,46 @@ export default function (
...
+            timeWindow: '1 minute',
+          },
+        },
+      },
+      async (req, res) => {
+        const { code, provider } = req.query;
+        try {
+          await licenseCodesService.isLicenseCodeAvailable(code, provider);
+          return res.status(200).send({ message: 'Code is available' });
+        } catch (err) {
```
sg-gs, line 285: Let's make it more clear: 

```
if (err instanceof LicenseCodeAlreadyAppliedError || err instanceof InvalidLicenseCodeError) {
   return res.status(404).send({ message: err.message });
}

req.log.error(`[LICENSE/CHECK/ERROR]: ${err.message}. STACK ${err.stack || 'NO STACK'}`);
return rep.status(500).send({ message: 'Internal Server Error' });
```

sg-gs: Think that if you have any error there, you want to know it. You cannot assume that all the errors are coming from what you expect to find. At that moment, you will need the logs to exist and to be easy to find to troubleshoot any issue @masterprog-cmd.

sg-gs (changes requested):

sg-gs (approved):

PR #37, opened 2023-09-06
[PB-706/PB-692]: fix/trial when cancelling plan

masterprog-cmd:
**PROBLEMS:**
The user is not charged the price he should be when using the free trial. We should add 3 months free and, once finished, charge the user for the plan without the 3-month discount.

If the user wants to upgrade his plan with the free trial, he will not be able to upgrade (either to a coupon code plan or to a normal plan).

**HOW TO SOLVE IT:**
If the user uses the free trial offer, we must set prorations_behavior to 'none'. If we want to upgrade a plan, we must add proration_behavior to 'create_proprations' and billing_cycle_behavior to 'now'. 

We have added trial_end: 'now' so that if the user upgrades his plan to have a free trial, he can do it without any problem.

**INFO:**
https://stripe.com/docs/billing/subscriptions/billing-cycle#prorating-changes-through-the-api

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -124,7 +124,11 @@ export default function (
         const { price_id: priceId, couponCode } = req.body;
 
         const user = await assertUser(req, rep);
-        await paymentService.updateSubscriptionPrice(user.customerId, priceId, couponCode);
+        await paymentService.updateSubscriptionPrice({
+          customerId: user.customerId,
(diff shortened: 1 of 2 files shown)


sg-gs:
This will affect the churn rate metric @masterprog-cmd

sg-gs (approved):

PR #39, opened 2023-09-11
[_]: fix/remove billing cycle param

masterprog-cmd:
**PROBLEM:**
When the user update the subscription, Stripe request for the 3D Secure again.

**FIX:**
I think the problem is `billing_cycle_anchor` because the user has to pay again, so let's remove that until Stripe give us a solution.

If the problem persists, I will look for another solution.


masterprog-cmd:
Do not merge yet.

PR #43, opened 2023-09-18
[PB-706/PB-692]: fix/payment issues

masterprog-cmd:
**PROBLEMS:**
* Cancellation flow (add 3 free months for the active subscription).
* Update/downgrade subscription does not work properly.

**FIX:**
* For the cancellation flow, the user will pay the total amount when the trial + subscription ends.
* To update/downgrade a subscription, the user will pay the total amount every time he changes the sub.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -194,7 +194,7 @@ export default function (
       } catch (err) {
         const error = err as Error;
         req.log.error(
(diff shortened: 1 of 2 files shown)


masterprog-cmd:
I have a problem with the prevent cancellation flow

sg-gs (approved):

PR #47, opened 2024-01-02
[_]: fix/compatibility-for-pricing-with-multiple-currencies

masterprog-cmd:
Get the correct price depending on the currency. 

Now, we have just one product with multi currency and we get the price depending on the currency we sent from the front end.

### DIFF
--- a/src/services/PaymentService.ts
+++ b/src/services/PaymentService.ts
@@ -345,7 +345,7 @@ export class PaymentService {
     };
   }
 
-  async getPrices(currency?: string): Promise<(DisplayPrice | undefined)[]> {
+  async getPrices(currency?: string): Promise<DisplayPrice[]> {
     const currencyValue = currency ?? 'eur';
     let priceProduct;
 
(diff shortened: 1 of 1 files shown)


src/services/PaymentService.ts
```diff
@@ -346,17 +346,19 @@ export class PaymentService {
...
       limit: 100,
     });
 
     return res.data
       .filter((price) => price.metadata.maxSpaceBytes)
       .map((price) => ({
         id: price.id,
-        currency: price.currency,
-        amount: price.unit_amount!,
+        currency: currencyValue,
```
sg-gs, line 360: This could have unexpected behaviors, at least assuming the below code makes sense. 

Just below you are doing: 
`amount: price.currency_options?.[currencyValue].unit_amount ?? price.unit_amount!`

I do not know in which case, the unit_amount is not set on the currency_options, but let's assume is like that (in which case, that value should be properly set from Stripe, instead of us patching it with a fallback). You could have the amount in $ (`price.unit_amount`) and the currency in € (`currencyValue` default value). 

I would enforce the `currency` value to match the `price.currency` to avoid this situation. I would remove that price from the listing and add a log for the situation so that in case we add a new price and the currency is not properly set, we instantly realize of that.

sg-gs (changes requested):

sg-gs (approved):

PR #59, opened 2024-04-29
[_]: feat/add method in webhook

masterprog-cmd:
Adding new method to store a new entry on the history of coupons used by a given user when `checkout.subscription.updated` webhook is triggered.

### DIFF
--- a/src/webhooks/handleCheckoutSessionCompleted.ts
+++ b/src/webhooks/handleCheckoutSessionCompleted.ts
@@ -74,8 +74,6 @@ export default async function handleCheckoutSessionCompleted(
     if (!(err instanceof CouponNotBeingTrackedError)) {
       log.error(`Error while adding user ${user.uuid} and coupon: `, err);
     }
-
-    throw err;
   }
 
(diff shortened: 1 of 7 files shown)


src/webhooks/handleCheckoutSessionCompleted.ts
```diff
@@ -59,12 +59,25 @@ export default async function handleCheckoutSessionCompleted(
...
+
+    const couponId = invoice.discount?.coupon.id;
+
+    if (couponId) {
+      await usersService.storeCouponUsedByUser(userData, couponId);
+    }
+  } catch (err) {
+    log.error(`Error while adding user ${user.uuid} and coupon`);
+
+    throw err;
```
sg-gs, line 74: Do not propagate, the `storeCouponUsedByUser` in fact throws an error if the coupon is not being tracked, but this could be intentional. In fact, you may log the error only if the error is not `instanceof` that error the function throws

sg-gs (changes requested):

PR #61, opened 2024-05-13
[PB-1860]: feat/call-workspace-initialization-through-drive-gateway

jzunigax2:
When `handleCheckoutSessionCompleted` hook is triggered an added check determines if the product redeemed is the B2B plan, in which case workspace initialization endpoint in drive-server-wip will be callled.

### New env variables

```
DRIVE_NEW_GATEWAY_URL=
DRIVE_NEW_GATEWAY_SECRET=
``` 
DRIVE_NEW_GATEWAY_URL: drive-server-wip's url ending with `/api`.
DRIVE_NEW_GATEWAY_SECRET: base 64 encoded string of the private key used to sign the JWT for requests to drive-server-wip, the public key should already be set there as `DRIVE_GATEWAY_PUBLIC_SECRET`.


src/webhooks/handleCheckoutSessionCompleted.ts
```diff
@@ -108,4 +110,15 @@ export default async function handleCheckoutSessionCompleted(
   } catch (err) {
     log.error(`Error in handleCheckoutSessionCompleted after trying to clear ${customer.id} subscription`);
   }
+
+  if (!productId) {
+    log.error(`Checkout session completed does not contain product id, customer: ${session.customer_email}`);
+    return;
+  }
+  
+  if (productId == PRODUCT_IDS.B2B) {
```
sg-gs, line 119: use strict match through `===`

src/constants.ts
```diff
@@ -1,2 +1,6 @@
 export const FREE_PLAN_BYTES_SPACE = 2 * 1024 * 1024 * 1024;
 export const FREE_INDIVIDUAL_TIER = 'free_individual_tier';
+
+export enum PRODUCT_IDS {
+  B2B = 'prod_Q6CYRA6w8fKZWK',
```
sg-gs, line 5: We can use the product metadata to determine if it is a workspace plan or not, rather than hardcoding IDs here, as it won't scale. An example of product metadata is the field `maxSpaceBytes`

.env.template
```diff
@@ -12,3 +12,5 @@ DRIVE_GATEWAY_PASSWORD=gatewaypass
 REDIS_HOST=payments-cache
 REDIS_PASSWORD=
 COMEBACK_COUPON_CODE=
+DRIVE_NEW_GATEWAY_URL=http://drive-server-wip:3004/api
+DRIVE_NEW_GATEWAY_SECRET=
```
sg-gs, line 16: Isn't there a value for this secret in the development environment? If not, you can suggest one and propose a PR on drive-server-wip also, so we keep this environment stable enough to run smoothly on local

sg-gs (changes requested):

src/constants.ts
jzunigax2: Thanks that's a much cleaner solution.

sg-gs (approved):

sg-gs:
Please @jzunigax2 set in the description the list of the new environment variables that need to be added so we can set the proper values in the production environment. 

**Do not merge until I confirm here by a comment that those values have been set in production**

jzunigax2:
> Please @jzunigax2 set in the description the list of the new environment variables that need to be added so we can set the proper values in the production environment.
> 
> **Do not merge until I confirm here by a comment that those values have been set in production**

@sg-gs done. Also, now that I think of it, these changes are related to B2B so should we follow a similar approach as in driver-server-wip where there is a dedicated branch holding all approved PRs related to B2B? Since I'll be working on more tasks related to this.

sg-gs:
Environment secrets set in the production cluster @apsantiso @jzunigax2

PR #62, opened 2024-05-21
[PB-2117]: feat/get active b2b subscriptions

jzunigax2:
Modify subscriptions endpoint to allow the retrieval of B2B subscriptions based on the metadata field on product `type="business"`. All products missing this metadata field are treated as if they where `"individual"` plans.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -154,37 +154,53 @@ export default function (
       return paymentService.getDefaultPaymentMethod(user.customerId);
     });
 
-    fastify.get('/subscriptions', async (req, rep) => {
-      let response: UserSubscription;
+    fastify.get<{
+      Querystring: { subscription_type?: 'B2B' };
+    }>(
+      '/subscriptions',
+      {
+        schema: {
+          querystring: {
+            type: 'object',
+            properties: { subscription_type: { type: 'string', enum: ['B2B'] } },
+          },
+        },
+      },
+      async (req, rep) => {
+        let response: UserSubscription;
 
-      const user: User = await assertUser(req, rep);
+        const user: User = await assertUser(req, rep);
+        const subscriptionType = req.query.subscription_type ?? 'individual';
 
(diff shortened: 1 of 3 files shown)


src/controller.ts
```diff
@@ -154,37 +154,53 @@ export default function (
       return paymentService.getDefaultPaymentMethod(user.customerId);
     });
 
-    fastify.get('/subscriptions', async (req, rep) => {
-      let response: UserSubscription;
+    fastify.get<{
+      Querystring: { subscription_type?: 'B2B' };
```
sg-gs, line 158: use camelCase, as we always do on JavaScript/TypeScript

src/controller.ts
```diff
@@ -154,37 +154,53 @@ export default function (
...
+    }>(
+      '/subscriptions',
+      {
+        schema: {
+          querystring: {
+            type: 'object',
+            properties: { subscription_type: { type: 'string', enum: ['B2B'] } },
```
sg-gs, line 165: Same here

src/services/PaymentService.ts
```diff
@@ -447,15 +478,31 @@ export class PaymentService {
   private async findIndividualActiveSubscription(customerId: CustomerId): Promise<Subscription> {
     const activeSubscriptions = await this.getActiveSubscriptions(customerId);
 
-    const individualActiveSubscription = activeSubscriptions.find(
-      (subscription) => subscription.items.data[0].price.metadata.is_teams !== '1',
-    );
+    const individualActiveSubscription = activeSubscriptions.find((subscription) => {
+      const isNotTeams = subscription.items.data[0].price.metadata.is_teams !== '1';
```
sg-gs, line 482: The teams thing is no longer needed, you can remove this line (this was the old b2b solution)

sg-gs (changes requested):

sg-gs (approved):

PR #64, opened 2024-05-29
[PB-2194]: feat/seat count handling on b2b subscriptions

jzunigax2:
When creating a checkout session if the product is B2B based on the metadata `type === 'business'` quantity becomes modifiable using Stripe's `adjustable_quantity` property, where `minimum` and `maximum` are set based on price metadata properties `minimumSeats` and `maximumSeats` respectively.

When handleCheckoutSessionHook is triggered quantity is taken from lineItems in order to call initializeWorkspace with `maxSpaceBytes: newStorageBytes * seats`.

### DIFF
--- a/src/webhooks/handleCheckoutSessionCompleted.ts
+++ b/src/webhooks/handleCheckoutSessionCompleted.ts
@@ -111,7 +111,8 @@ export default async function handleCheckoutSessionCompleted(
   }
 
   if (product.metadata?.type === 'business') {
-    const address = customer.address?.line1 || undefined;
-    await usersService.initializeWorkspace(user.uuid, Number(maxSpaceBytes), address);
+    const amountOfSeats = lineItems.data[0]!.quantity ?? parseInt(price.metadata.minimumSeats);
+    const address = customer.address?.line1 ?? undefined;
+    await usersService.initializeWorkspace(user.uuid, Number(maxSpaceBytes), amountOfSeats, address);
   }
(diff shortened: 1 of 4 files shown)


src/webhooks/handleCheckoutSessionCompleted.ts
```diff
@@ -111,7 +111,8 @@ export default async function handleCheckoutSessionCompleted(
   }
 
   if (product.metadata?.type === 'business') {
-    const address = customer.address?.line1 || undefined;
-    await usersService.initializeWorkspace(user.uuid, Number(maxSpaceBytes), address);
+    const amountOfSeats = lineItems.data[0]!.quantity ?? parseInt(price.metadata.minimumSeats);
```
sg-gs, line 114: Is this the value the customer selects when picking the quantity of users for this plan? This seems like information related to min/max seats per product rather than the customer's picked value

sg-gs (changes requested):

src/webhooks/handleCheckoutSessionCompleted.ts
jzunigax2: It is the amount the user selected/modified during checkout, I added `?? parseInt(price.metadata.minimumSeats);` in case that came out as undefined but now that I think of it that, should never happen. I'll change this to a non null assertion.

sg-gs (approved):

PR #65, opened 2024-06-07
[MKT-305]: feat/handle checkout manually

masterprog-cmd:
Endpoints created to manage the payments manually.

The endpoints are the following ones:
- Create a customer and return de customerId.
- Create a subscription and return de client_secret to the client side.
- Create a Payment Intent (for one_time payments) ad return the client_secret to the client side.
- Handle the event received in the webhook controller (payment_intent.succeeded) for both cases.


src/controller.ts
```diff
@@ -71,6 +79,116 @@ export default function (
       }
     });
 
+    fastify.post<{ Body: { name: string; email: string } }>(
```
sg-gs, line 82: Rate limit this to 5 per hour.

src/controller.ts
```diff
@@ -71,6 +79,116 @@ export default function (
...
+      const { email } = req.query;
+
+      if (!email) {
+        return res.status(404).send({
+          message: 'Email should be provided',
+        });
+      }
+
+      try {
+        const { id } = await paymentService.getCustomerIdByEmail(email);
```
sg-gs, line 137: Validate that the user is not found on `getCustomerIdByEmail`. If the error is not a custom one or expected to happen, send a 500 and log the error

src/controller.ts
```diff
@@ -71,6 +79,116 @@ export default function (
...
+              },
+            },
+          },
+        },
+      },
+      async (req, res) => {
+        const { customerId, priceId, promoCodeId } = req.body;
+
+        try {
+          const subscriptionSetUp = await paymentService.createSubscription(customerId, priceId, promoCodeId);
```
sg-gs, line 173: Send a signed token on `createCustomer` with the customerId inside so we can ensure that the user doing the request owns the customer id that is provided to subscribe

src/controller.ts
```diff
@@ -230,6 +348,45 @@ export default function (
...
+          properties: {
+            customerId: { type: 'string' };
+            planId: { type: 'string' };
+            amount: { type: 'number' };
+            promoCodeName: { type: 'string' };
+          };
+        };
+      };
+    }>('/payment-intent', async (req, rep) => {
+      const { customerId, amount, planId, promoCodeName } = req.query;
```
sg-gs, line 370: Use the signed token here also, to avoid impersonation issues

src/services/PaymentService.ts
```diff
@@ -100,6 +123,89 @@ export class PaymentService {
...
+    customerId: CustomerId,
+    amount: number,
+    planId: string,
+    promoCodeName?: string,
+  ): Promise<PaymentIntent> {
+    if (!customerId || !amount || !planId) {
+      throw new MissingParametersError(['customerId', 'amount', 'planId']);
+    }
+
+    const product = await this.provider.prices.retrieve(planId);
```
sg-gs, line 178: Use `priceId` instead of `planId`

src/webhooks/handlePaymentIntentCompleted.ts
```diff
@@ -0,0 +1,137 @@
...
+      error.stack ?? error.message,
+    );
+
+    throw err;
+  }
+
+  try {
+    const customerByUuid = await usersService.findUserByUuid(user.uuid);
+
+    // if (!customerByUuid) throw Error();
```
sg-gs, line 99: Remove comment

src/webhooks/handleSubscriptionCanceled.ts
```diff
@@ -32,15 +33,19 @@ export default async function handleSubscriptionCanceled(
   if (productType === UserType.Business) {
     return usersService.destroyWorkspace(uuid);
   }
- 
+
+  if (hasBoughtALifetime) {
+    // This user has purchased a lifetime plan by having a subscription.
```
sg-gs, line 38: _This user has switched from a subscription to a lifetime, therefore we do not want to downgrade his space_

sg-gs (approved):

PR #68, opened 2024-06-25
[PB-2115] feat: b2b update subscription payment method

evillalba94:
## Updates
- Add/Update webhooks to update the payment method according to the `userType` (`individual`, `business`)
- Modify `setup-intent` to add the `metadata` field. From the frontend, they need to send the appropriate `userType` to handle the `setup_intent.succeeded` event and update the appropriate subscription.
- Update the workspace storage if the `userType` is `business`

## Related to
- SDK : https://github.com/internxt/sdk/pull/216

### DIFF
--- a/src/webhooks/handleSetupIntentSucceded.ts
+++ b/src/webhooks/handleSetupIntentSucceded.ts
@@ -1,0 +1,20 @@
+import Stripe from 'stripe';
+import { PaymentService } from '../services/PaymentService';
+
+type SubscriptionType = 'individual' | 'business';
+
+export default async function handleSetupIntentSucceded(
+  setupIntent: Stripe.SetupIntent,
+  paymentService: PaymentService,
+): Promise<void> {
+  const customerId = setupIntent.customer as string;
+  const paymentMethodId = setupIntent.payment_method as string;
+  const setupIntentMetdata = setupIntent.metadata as Stripe.Metadata;
+  
+  const subscriptionType: SubscriptionType = setupIntentMetdata?.subscriptionType as SubscriptionType;
+  const setupSubscriptionPayment = subscriptionType === 'business' || subscriptionType === 'individual';
+
+  if (setupSubscriptionPayment) {
+    await paymentService.updateSubscriptionPaymentMethod(customerId, paymentMethodId, subscriptionType);
+  }
+}
--- a/src/webhooks/handleSetupIntentSucceded.ts
+++ b/src/webhooks/handleSetupIntentSucceded.ts
@@ -1,7 +1,6 @@
 import Stripe from 'stripe';
 import { PaymentService } from '../services/PaymentService';
-
-type SubscriptionType = 'individual' | 'business';
+import { UserType } from '../core/users/User';
 
 export default async function handleSetupIntentSucceded(
   setupIntent: Stripe.SetupIntent,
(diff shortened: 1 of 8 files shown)


evillalba94:
@sg-gs Could you review this? thanks

evillalba94:
@sg-gs  Could you review this? Lmk

src/webhooks/index.ts
```diff
@@ -58,18 +58,11 @@ export default function (
             usersService,
             event.data.object as Stripe.Subscription,
             cacheService,
+            paymentService,
             fastify.log,
             config,
           );
           break;
-        case 'payment_method.attached':
```
sg-gs, line 65: Why are we removing this webhook? 

According to [Stripe's documentation](https://docs.stripe.com/api/events/types) this event is triggered when a new payment method is attached to a customer (which could happen on the first payment). Therefore, if we remove this, how can we ensure that the payment method is saved the first time?

src/webhooks/handleSubscriptionUpdated.ts
```diff
@@ -20,20 +22,38 @@ export default async function handleSubscriptionUpdated(
...
+  const productType = productMetadata?.type === UserType.Business ? UserType.Business : UserType.Individual;
 
   try {
-    await cacheService.clearSubscription(customerId);
+    await cacheService.clearSubscription(customerId, productType);
   } catch (err) {
     log.error(`Error in handleSubscriptionUpdated after trying to clear ${customerId} subscription`);
   }
 
+  if (productType == UserType.Business) {
```
sg-gs, line 36: Use the strict match (`===`) always

src/webhooks/handleSetupIntentSucceded.ts
```diff
@@ -0,0 +1,18 @@
...
+  paymentService: PaymentService,
+): Promise<void> {
+  const customerId = setupIntent.customer as string;
+  const paymentMethodId = setupIntent.payment_method as string;
+  const setupIntentMetadata = setupIntent.metadata as Stripe.Metadata;
+  
+  const userType = setupIntentMetadata?.userType as UserType;
+
+  if ([UserType.Individual, UserType.Business].includes(userType)) {
+    await paymentService.updateSubscriptionPaymentMethod(customerId, paymentMethodId, userType);
```
sg-gs, line 16: Are we sure that this [event](https://docs.stripe.com/api/setup_intents) is reliable enough to assume the payment is collected, thus, updating the payment method?

sg-gs (changes requested):

src/webhooks/handleSetupIntentSucceded.ts
masterprog-cmd: I think this event is enough for that @sg-gs. The event is triggered when the setup intent has been completed: https://docs.stripe.com/api/events/types#event_types-setup_intent.succeeded. 

From what I understand now, when creating a setup intent you get the payment method id which is what we use to update the payment method.

src/webhooks/index.ts
evillalba94: We are removing this because we will now handle cases that come through this event: `setup_intent.succeeded` (a few lines below in the code). Since through `/setup-intent` we will be sending the `userType` (`individual` or `business`) in the metadata to update the corresponding subscription.

src/webhooks/handleSetupIntentSucceded.ts
evillalba94: @sg-gs 
Thanks for answering @masterprog-cmd . Yes, correct, currently when we generate a Stripe Setup Intent (`/setup-intent`) we will be adding the `userType` in the metadata to know which subscription to work with when receiving a setupIntent event.

If there is no matching `userType` in the metadata, it means that the call for this event was made in another way and not from our endpoint side through the front end.

Once identified, we will then update the subscription of the corresponding user.

src/webhooks/handleSubscriptionUpdated.ts
```diff
@@ -20,20 +22,38 @@ export default async function handleSubscriptionUpdated(
...
+    const customer = await paymentService.getCustomer(customerId);
+    if (customer.deleted) {
+      log.error(
+        `Customer object could not be retrieved in subscription updated handler with id ${customer.id}`,
+      );
+      return;
+    }
+    const { maxSpaceBytes: priceMaxSpaceBytes } = subscription.items.data[0].price.metadata as PriceMetadata;
+    const amountOfSeats = subscription.items.data[0]!.quantity!;
+
```
sg-gs, line 46: What happens if the subscription is canceled in the Business plan case? It does not seem to be handled right now

sg-gs (approved):

src/webhooks/index.ts
masterprog-cmd: But I think `setup_intent.succeeded` is not triggered at checkout. 

This is the event we use to save the user's payment method once the payment is completed if I'm not mistaken. Have you checked that the card is saved by making a payment from the checkout?
@evillalba94

evillalba94: In that case, we handle the payment at `checkout.session.completed`. The payment is automatically linked to the subscription.

PR #71, opened 2024-07-08
Chore: handle user in webhooks

evillalba94:
### Update
- Getting the `user` on webhooks. The user must be captured to continue with the operations (**update storage**, **space**, etc), when paying for a `business subscription` it is difficult to obtain the user since it has not yet been created if there is NOT a `subscription individual/personal` before, so it is necessary to verify that the user exists but rather create it.

- To create it, the `user.uuid` is required, however, if the user is not registered in the **payments database** then we will have to search for it on the `drive-server-wip`. For this reason, a **New Endpoint** was created in `drive-server-wip` that allows us to search for the user through the email provided by `Stripe.Customer`.

### PR related
- https://github.com/internxt/drive-server-wip/pull/354

### DIFF
--- a/src/webhooks/handleSubscriptionUpdated.ts
+++ b/src/webhooks/handleSubscriptionUpdated.ts
@@ -3,7 +3,7 @@ import Stripe from 'stripe';
 import { FREE_INDIVIDUAL_TIER, FREE_PLAN_BYTES_SPACE } from '../constants';
 import CacheService from '../services/CacheService';
 import { PaymentService, PriceMetadata } from '../services/PaymentService';
-import { findUserByEmail, StorageService, updateUserTier } from '../services/StorageService';
+import { StorageService, updateUserTier } from '../services/StorageService';
 import { UsersService } from '../services/UsersService';
 import { AppConfig } from '../config';
 import { User, UserType } from '../core/users/User';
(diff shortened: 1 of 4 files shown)


sg-gs:
As we are integrating this change on the branch that will be released together with Drive's Workspaces, why are we still using the old gateway? drive-server is a deprecated project and we will avoid making changes there @evillalba94

src/webhooks/handleSubscriptionUpdated.ts
sg-gs, line 1: I can accept the changes on the checkout session completed, but not on the subscription updated. If the user has a subscription, it should exist on this service and is a mandatory condition. Otherwise, we should fix that anomaly in other ways like the support service, which is out of the scope of the payments service, therefore I would just remove the changes done here.

sg-gs (changes requested):

evillalba94:
The implementation will not take place. If necessary, it will be carried out in the most current drive project.

> As we are integrating this change on the branch that will be released together with Drive's Workspaces, why are we still using the old gateway? drive-server is a deprecated project and we will avoid making changes there @evillalba94

Done! I made the changes to use `drive-server-wip`. PR related: https://github.com/internxt/drive-server-wip/pull/354

sg-gs (approved):

PR #81, opened 2024-07-29
[MKT-408]: feat/use promotion code ID instead of coupon ID

masterprog-cmd:
## Context
In production, the coupon ID is used to create the session with the discount automatically added. Now, the promotion code ID is used to do so.

## Description
Using the promotion code Id and not the coupon Id gives us the advantage of being able to use disposable promo codes. That is, we can configure them with an expiration date to be unusable, so we can use the promotion code (the same code that is added in the checkout when the user must add it manually). This allows us to have more flexibility when redeeming coupons.

## Changes in the codebase
Now, the promotion code ID and not the coupon ID is used when creating the checkout. And because of this, an endpoint has been enabled to get the promotion code information by name, where you get the id, the percentage/fixed price that is discounted.

The same flow is maintained as it is now. The coupon information is obtained from the website, where the id will be sent to drive-web to create the checkout session with this promotion code id, and where finally the user will be redirected to the Stripe checkout.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -396,7 +396,6 @@ export default function (
       } catch (error) {
         const err = error as Error;
         if (err instanceof NotFoundPromoCodeByNameError) {
-          console.log('SI');
           return rep.status(404).send({ message: err.message });
         }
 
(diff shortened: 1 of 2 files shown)


masterprog-cmd:
⚠️ Do not merge it before merging the website / drive-web PRs.

sg-gs (approved):

PR #82, opened 2024-07-30
[PB-2387]: feat/update subscriptions handle individual and business plans

jzunigax2:
Allow the upgrade or downgrade of b2b subscription, mantaining the number of seats from the current active plan.

### DIFF
--- a/src/services/PaymentService.ts
+++ b/src/services/PaymentService.ts
@@ -167,7 +167,7 @@ export class PaymentService {
    * @param additionalOptions - Additional options to update the subscription (all the options from Stripe.SubscriptionUpdateParams)
    * @returns The updated subscription
    */
-  async updateSub({
+  async updateIndividualSub({
     customerId,
     priceId,
     additionalOptions,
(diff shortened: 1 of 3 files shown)


src/services/UsersService.ts
```diff
@@ -193,14 +193,17 @@ export class UsersService {
         'Content-Type': 'application/json',
         Authorization: `Bearer ${jwt}`,
       },
-      data: {
+    };
+
+    await this.axios.put(
```
sg-gs, line 198: Is it the same as sending the body as the second parameter rather than sending it as the data field of the configuration?

src/services/PaymentService.ts
```diff
@@ -178,6 +178,16 @@ export class PaymentService {
     additionalOptions?: Partial<Stripe.SubscriptionUpdateParams>;
   }) {
     const individualActiveSubscription = await this.findIndividualActiveSubscription(customerId);
+
+    const newPrice = await this.provider.prices.retrieve(priceId, {
```
sg-gs, line 182: Why is this needed?

sg-gs (changes requested):

src/services/UsersService.ts
jzunigax2: I would have guessed it's the same, but for some odd reason when setting the body as data in the configurations the headers where lost. Resulting in 401 responses since the jwt was not received.

src/services/PaymentService.ts
jzunigax2: It's only purpose would be to validate the priceId actually belongs to an individual plan. While testing I accidentally sent a b2b priceId to an individual plan update and it basically resulted in a "lost" individual sub. I got downgraded to free. I guess this shouldn't really be an issue since the frontend wouln't specify an incorrect priceId?

sg-gs: It's fine to validate these things, it's okay then

sg-gs (dismissed):

PR #86, opened 2024-08-06
[MKT-413]: fix/coupon codes

masterprog-cmd:
Check if the coupon can be used for each case:
- If it is for first purchase only.
- If it is for a specific plan.

Also a try catch has been added to the function that handles the customer.subscription.updated event so that it doesn't always pop a 500.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -647,7 +647,7 @@ export default function (
       } catch (error) {
         const err = error as Error;
         if (err instanceof NotFoundPlanByIdError) {
-          return rep.status(404).send({ message: err.message });
+          return rep.status(404).send(err.message);
         }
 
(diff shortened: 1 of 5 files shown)


inxtci:
> [!CAUTION]
> Preview failed: Not deployed after 5 minutes check 💥
> **Last status received: Progressing**

inxtci:
> [!CAUTION]
> Preview failed: Not deployed after 5 minutes check 💥
> **Last status received: Degraded**

inxtci:
> [!CAUTION]
> Preview failed: Not deployed after 5 minutes check 💥
> **Last status received: Degraded**

inxtci:
> [!CAUTION]
> Preview failed: Not deployed after 5 minutes check 💥
> **Last status received: Degraded**

sg-gs (approved):

PR #110, opened 2024-08-27
[_]: fix/send error messages when creating subscription endpoint fails

masterprog-cmd:
send error messages to find out what is going wrong in the integrated checkout flow.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -289,7 +289,7 @@ export default function (
         const { customerId, priceId, currency, token, promoCodeId } = req.body;
 
         if (!customerId || !priceId) {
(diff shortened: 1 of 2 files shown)


src/controller.ts
```diff
@@ -288,6 +288,10 @@ export default function (
       async (req, res) => {
         const { customerId, priceId, currency, token, promoCodeId } = req.body;
 
+        if (customerId === undefined || priceId === undefined) {
```
sg-gs, line 291: Why not using `!` instead of `=== undefined` ?

sg-gs (changes requested):

PR #119, opened 2024-11-12
[_]: fix/avoid canceling business sub, avoid updating the lifetime field when it is not needed and return the correct sub for each user type

xabg2:
- Return the correct plan for both individual and subscriptions, because when the individual sub is lifetime, `type=lifetime` is returned for both subs (individual and business).
- Prevent cancellation of business subscriptions when the product is an individual lifetime. Now we filter the product and check if is for individuals before canceling it.
- Prevent lifetime field update if the sub is not individual.

### DIFF
--- a/src/webhooks/handleInvoiceCompleted.ts
+++ b/src/webhooks/handleInvoiceCompleted.ts
@@ -102,13 +102,13 @@ export default async function handleInvoiceCompleted(
 
   try {
     const userActiveSubscription = await paymentService.getActiveSubscriptions(customer.id);
-    const hasActiveSubscription = userActiveSubscription.length > 0;
     const hasIndividualActiveSubscription = userActiveSubscription.find(
       (sub) =>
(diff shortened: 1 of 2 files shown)


src/webhooks/handleInvoiceCompleted.ts
```diff
@@ -103,9 +103,13 @@ export default async function handleInvoiceCompleted(
   try {
     const userActiveSubscription = await paymentService.getActiveSubscriptions(customer.id);
     const hasActiveSubscription = userActiveSubscription.length > 0;
```
sg-gs, line 105: Due to the changes, this check is no longer needed, as the case `hasIndividualActiveSubscription` is more specific than the `hasActiveSubscription` one

sg-gs (changes requested):

src/webhooks/handleInvoiceCompleted.ts
xabg2: Done

sg-gs (approved):

PR #135, opened 2024-11-28
[_]: feat/users tests coverage

xabg2:
Adding tests for `users.service.ts` and enabling the Sonarcloud coverage.


src/config.ts
```diff
@@ -46,11 +45,9 @@ const variablesToCheck = [
   ...(process.env.NODE_ENV === 'production' ? mandatoryVariablesOnlyInProd : []),
 ];
 
-
-
 const undefinedMandatoryVariables = variablesToCheck.filter((key) => !process.env[key]);
 
-if (undefinedMandatoryVariables.length) {
+if (!isTestEnv && undefinedMandatoryVariables.length) {
```
sg-gs, line 50: The test environment will still require the environment variables to be present. This seems like a bad idea.

tests/src/services/users.service.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    });
+  });
+
+  describe('Find customer by Customer ID', () => {
+    it('should find a user by customerId successfully', async () => {
+      (usersRepository.findUserByCustomerId as jest.Mock).mockResolvedValue(mocks.mockedUser);
+
+      const result = await usersService.findUserByCustomerID(mocks.mockedUser.customerId);
+
+      expect(result).toEqual(mocks.mockedUser);
```
sg-gs, line 82: Use `strictEqual` always if possible

tests/src/services/users.service.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    });
+  });
+
+  describe('Find customer by User UUId', () => {
+    it('should find a user by UUID successfully', async () => {
+      (usersRepository.findUserByUuid as jest.Mock).mockResolvedValue(mocks.mockedUser);
+
+      const result = await usersService.findUserByUuid(mocks.mockedUser.uuid);
+
+      expect(result).toEqual(mocks.mockedUser);
```
sg-gs, line 103: Use `strictEqual` always if possible

tests/src/services/users.service.test.ts
```diff
@@ -0,0 +1,235 @@
...
+              (sub) => sub.product?.metadata.type !== 'business',
+            ) as unknown as ExtendedSubscription[],
+          ),
+        );
+      const cancelSubscriptionSpy = jest.spyOn(paymentService, 'cancelSubscription').mockImplementation(voidPromise);
+      const changeStorageSpy = jest.spyOn(storageService, 'changeStorage').mockImplementation(voidPromise);
+
+      await paymentService.getActiveSubscriptions(mocks.mockedUser.customerId);
+
+      await usersService.cancelUserIndividualSubscriptions(mocks.mockedUser.customerId);
```
sg-gs, line 134: Why are you calling multiple methods in the same test?

tests/src/services/users.service.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        .mockImplementation(() => Promise.resolve(mocks.mockActiveSubscriptions as unknown as ExtendedSubscription[]));
+
+      const cancelSubscriptionSpy = jest.spyOn(paymentService, 'cancelSubscription').mockImplementation(voidPromise);
+
+      const changeStorageSpy = jest.spyOn(storageService, 'changeStorage').mockImplementation(voidPromise);
+
+      await usersService.cancelUserB2BSuscriptions(mocks.mockedUser.customerId);
+      await storageService.changeStorage(mocks.mockedUser.uuid, FREE_PLAN_BYTES_SPACE);
+
+      const b2bSubscriptions = mocks.mockActiveSubscriptions.filter((sub) => sub.product?.metadata.type === 'business');
```
sg-gs, line 158: It's better to separate between business and individual subscriptions in the mocks rather than using logic (like filtering) in a test, as any kind of logic should be tested.

tests/src/services/users.service.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    });
+
+    it('should throw CouponNotBeingTrackedError when the coupon is not tracked', async () => {
+      (couponsRepository.findByCode as jest.Mock).mockResolvedValue(null);
+
+      await expect(usersService.storeCouponUsedByUser(mocks.mockedUser, 'INVALID_COUPON')).rejects.toThrow(
+        CouponNotBeingTrackedError,
+      );
+
+      expect(couponsRepository.findByCode).toHaveBeenCalledWith('INVALID_COUPON');
```
sg-gs, line 191: Storing the `INVALID_COUPON` string in a constant is better for test maintenance purposes

tests/src/services/users.service.test.ts
sg-gs, line 1: Use `When {something} then {something}` rather than `should {something}` as it enforces high-level thinking. Please use high-level speaking so the tests are easy to understand on a CI pipeline if they fail. For instance, instead of: 

_should find a user by customerId successfully_

use 

_When looking for a customer by its ID with the correct params, then the customer is found_

sg-gs (changes requested):

sg-gs (approved):

sg-gs:
The coverage seems to not be working properly @xabg2

PR #140, opened 2024-12-02
[_]: feat/prepare-crypto-payments

sg-gs:
This PR generates and activates an invoice so that it can be paid using cryptocurrencies if the user selects one that is available. 

In addition, a link to a QR code is generated, which the user can use to easily pay from their wallet. The URI will also be provided so that the client can make the QR code clickable for desktop users if they want to pay from their computer.

⚠️ To generate the QR code, a couple of  new **ENVIRONMENT VARIABLE** have been added: 
- `CHART_API_URL`
- `DRIVE_WEB_URL`

We need to add them to production before deployment.


src/services/payment.service.ts
```diff
@@ -406,16 +449,50 @@ export class PaymentService {
...
+
+    if (isLifetime && this.bit2MeService.isAllowedCurrency(currency)) {
+      const invoice = await this.bit2MeService.createCryptoInvoice({
+        description: `Payment for lifetime product ${priceId}`,
+        priceAmount: invoiceItem.amount,
+        priceCurrency: currency,
+        title: `Invoice from Stripe ${finalizedInvoice.id}`,
+        securityToken: jwt.sign(
+          {
+            invoiceId: finalizedInvoice.id,
```
xabg2, line 467: Perhaps we could call it `stripeInvoiceId` to differentiate this invoice ID which is from Stripe, from the bit2me invoice ID.

xabg2:
Hey @sg-gs , this PR is ready for review.

src/services/payment.service.ts
```diff
@@ -406,16 +447,52 @@ export class PaymentService {
...
+        title: `Invoice from Stripe ${finalizedInvoice.id}`,
+        securityToken: jwt.sign(
+          {
+            stripeInvoiceId: finalizedInvoice.id,
+            customerId: customerId,
+          },
+          config.JWT_SECRET,
+        ),
+        foreignId: finalizedInvoice.id,
+        cancelUrl: 'https://drive.internxt.com/cancel',
```
sg-gs, line 471: Isn't this set as an environment variable? It should be

src/services/payment.service.ts
```diff
@@ -406,16 +447,52 @@ export class PaymentService {
...
+        successUrl: 'https://drive.internxt.com/success',
```
sg-gs, line 472: Same here

src/services/payment.service.ts
```diff
@@ -406,16 +447,52 @@ export class PaymentService {
...
+        purchaserEmail: userEmail || 'hello@internxt.com',
```
sg-gs, line 473: If this is mandatory, then it should not have a default value. If not, it's pretty strange, unless there is an underlying reason I do not know

src/config.ts
```diff
@@ -22,6 +22,7 @@ const mandatoryVariables = [
   'CRYPTO_PAYMENTS_PROCESSOR_API_KEY',
   'VPN_URL',
   'PC_CLOUD_TRIAL_CODE',
+  'CHART_API_URL',
```
sg-gs, line 25: If this is added, add it to the .env.template also

src/services/payment.service.ts
xabg2: Agree.

xabg2:
Hey @sg-gs , I think it is ready for Review again. I made the requested changes.

sg-gs (commented):
LGTM @xabg2

sg-gs:
Environment variables have been set @xabg2

PR #142, opened 2024-12-03
[_]: feat/handle lost disputes through the webhook

xabg2:
When a dispute is closed with status lost, if it is for a subscription, we will cancel it and if it is a lifetime, the lifetime status will be set to false. 

For both cases the storage will be changed to 1GB (free plan).

This PR also contains the update of a type that has become obsolete and the correction of a grammatical problem of a function.

### DIFF
--- a/src/webhooks/handleDisputeLost.ts
+++ b/src/webhooks/handleDisputeLost.ts
@@ -1,0 +1,52 @@
+import Stripe from 'stripe';
+import { UsersService } from '../services/users.service';
+import { StorageService } from '../services/storage.service';
+import CacheService from '../services/cache.service';
+import { AppConfig } from '../config';
+import handleLifetimeRefunded from './handleLifetimeRefunded';
+import { PaymentService } from '../services/payment.service';
+import { FastifyBaseLogger } from 'fastify';
+
+interface HandleDisputeLostProps {
+  charge: Stripe.Dispute;
+  stripe: Stripe;
+  paymentService: PaymentService;
+  usersService: UsersService;
+  storageService: StorageService;
+  cacheService: CacheService;
+  log: FastifyBaseLogger;
+  config: AppConfig;
+}
+
+export async function handleDisputeLost({
+  charge,
+  stripe,
+  paymentService,
+  usersService,
+  storageService,
+  cacheService,
+  log,
+  config,
+}: HandleDisputeLostProps) {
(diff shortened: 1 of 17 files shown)


src/webhooks/handleInvoicePaymentFailed.ts
```diff
@@ -22,7 +22,7 @@ export default async function handleInvoicePaymentFailed(
   invoice: Stripe.Invoice,
   objectStorageService: ObjectStorageService,
   paymentService: PaymentService,
-  logger: FastifyLoggerInstance,
+  logger: FastifyBaseLogger,
```
sg-gs, line 25: Why this change is needed?

xabg2: The change is not needed for this PR, but FastifyLogerInstance is deprecated, so I updated it to the new one.

src/webhooks/handleDisputeLost.ts
```diff
@@ -0,0 +1,52 @@
...
+  stripe: Stripe;
+  paymentService: PaymentService;
+  usersService: UsersService;
+  storageService: StorageService;
+  cacheService: CacheService;
+  log: FastifyBaseLogger;
+  config: AppConfig;
+}
+
+export async function handleDisputeLost({
```
sg-gs, line 21: `handleDisputeResult` could be a better name, as you do not know if the dispute has been lost or not, for what I see in the below checks

src/webhooks/handleDisputeLost.ts
```diff
@@ -0,0 +1,52 @@
...
+  paymentService,
+  usersService,
+  storageService,
+  cacheService,
+  log,
+  config,
+}: HandleDisputeLostProps) {
+  const isDisputeLost = charge.status === 'lost';
+  const chargeId = charge.charge as string;
+
```
sg-gs, line 33: Return if the dispute is not lost, so you reduce the cognitive complexity of the if that wraps the entire `try-catch`

sg-gs (changes requested):

src/webhooks/handleInvoicePaymentFailed.ts
sg-gs: Okay, add the reference link of the Fastify documentation that mentions the deprecation in the description of this PR so we keep track of that in case anything happens.

PR #144, opened 2024-12-12
[_]: feat/check what order ids has been refunded

xabg2:
Check which orders (payment attempts) have been refunded with your ID.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   "dependencies": {
     "@fastify/cors": "^10.0.1",
     "@fastify/jwt": "^9.0.1",
+    "@fastify/multipart": "^9.0.1",
     "@fastify/rate-limit": "^10.1.1",
     "axios": "^1.7.7",
     "dayjs": "^1.11.13",
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,9 +1,11 @@
 import Stripe from 'stripe';
 import fastifyCors from '@fastify/cors';
 import Fastify, { FastifyInstance } from 'fastify';
+import multipart from '@fastify/multipart';
 import { AppConfig } from './config';
 import controller from './controller/payments.controller';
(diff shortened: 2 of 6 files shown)


src/controller/gateway.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+    fastify.addHook('onRequest', async (request, reply) => {
+      try {
+        await request.jwtVerify();
+      } catch (err) {
+        reply.status(401).send({ error: 'Unauthorized: Invalid token' });
+      }
+    });
+    fastify.post('/check-order-id', async (req, res) => {
+      try {
+        const file = await req.file();
```
sg-gs, line 32: You need to control the size, otherwise a big file could take down the replica of the server if it refills the entire RAM

sg-gs: Just check the size in number and enforce < 10MB

src/services/orders.service.ts
```diff
@@ -0,0 +1,74 @@
...
+      type: 'payment_intent',
+      refunded,
+    };
+  } catch (error) {
+    log.error(`Error processing order ID ${orderId}: ${(error as Error).message}`);
+    throw new InternalServerError(`Failed to process order ID ${orderId}`);
+  }
+}
+
+export async function processUploadedFile(
```
sg-gs, line 39: Add a JSDoc specifying which is the expect file format so anyone can be referred to that explanation without having to guess how this function handles it

src/config.ts
sg-gs, line 1: If you have a new environment variable, and specially if it is mandatory, add a warning to the description of this Pull Request, otherwise we may forget that detail and break the entire service by forgetting to set the variable

sg-gs (changes requested):

src/config.ts
xabg2: Done!

PR #150, opened 2025-01-17
[PB-3639]: fix/get existing customer

xabg2:
Using the uuid to get the existing user from our DB if it exists before creating a new one in Stripe when calling `POST /create-customer`. Also, we get the user by uuid when payment invoices are completed to update it in the DB if needed.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -29,7 +29,13 @@ import { Bit2MeService } from './services/bit2me.service';
 import { TiersService } from './services/tiers.service';
 import { MongoDBTiersRepository, TiersRepository } from './core/users/MongoDBTiersRepository';
 
-const start = async (mongoTestClient?: MongoClient): Promise<FastifyInstance> => {
+const start = async (
+  mongoTestClient?: MongoClient,
+  mocksServices?: {
+    usersService: UsersService;
+    paymentsService: PaymentService;
+  },
+): Promise<FastifyInstance> => {
   const mongoClient = mongoTestClient ?? (await new MongoClient(envVariablesConfig.MONGO_URI).connect());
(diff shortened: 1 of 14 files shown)


src/server.ts
```diff
@@ -29,7 +29,13 @@ import { Bit2MeService } from './services/bit2me.service';
 import { TiersService } from './services/tiers.service';
 import { MongoDBTiersRepository, TiersRepository } from './core/users/MongoDBTiersRepository';
 
-const start = async (mongoTestClient?: MongoClient): Promise<FastifyInstance> => {
+const start = async (
+  mongoTestClient?: MongoClient,
+  mocksServices?: {
```
sg-gs, line 34: This a very bad idea. Tests are for covering the real codebase, you cannot mix both things. What you need to do here involves mechanisms like sinon's [stub](https://sinonjs.org/releases/latest/stubs/) method

tests/src/mocks/index.ts
```diff
@@ -9,6 +9,16 @@ import { Tier } from '../../../src/core/users/MongoDBTiersRepository';
 const randomDataGenerator = new Chance();
 
 export default function getMocks() {
+  const user = {
```
sg-gs, line 12: If you can use functions to instantiate mocks instead of having to create a const each time, that would be cleaner, easier to use and faster to work with. Example [here](https://github.com/internxt/drive-server-wip/blob/8ed79deac574946f542fe9b9f0b904a864c3864c/test/fixtures.ts#L171) where you can instantiate a user, by default it has as real as possible fields (using chance). It has of course things to improve, but it's far more scalable than this way of creating mocks.

sg-gs (changes requested):

PR #154, opened 2025-02-13
[_]: tests/use functions for mocks instead

xabg2:
This PR refactors the way mocks are created and utilized in test cases. Previously, mocks were defined as constant variables, which limited flexibility and could lead to unintended state persistence across multiple tests.

With this update, mocks are now encapsulated within functions, ensuring that each test case receives a fresh and isolated mock instance when needed. This approach enhances test reliability, reduces potential side effects, and improves maintainability by providing greater control over mock initialization.

By implementing this change, we improve test consistency and make the mocking strategy more scalable for future test expansions.

### DIFF
--- a/tests/src/webhooks/handleDisputeResult.test.ts
+++ b/tests/src/webhooks/handleDisputeResult.test.ts
@@ -40,7 +40,7 @@ jest.mock('stripe', () => {
         retrieve: jest.fn(),
       },
       subscriptions: {
-        cancel: jest.fn(), // <-- ¡agrega esto!
+        cancel: jest.fn(),
       },
     })),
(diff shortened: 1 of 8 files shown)


tests/src/webhooks/handleDisputeResult.test.ts
```diff
@@ -31,6 +39,9 @@ jest.mock('stripe', () => {
       invoices: {
         retrieve: jest.fn(),
       },
+      subscriptions: {
+        cancel: jest.fn(), // <-- ¡agrega esto!
```
sg-gs, line 43: Remove this comment

PR #164, opened 2025-02-17
[PB-3884]: feat/retrieve the active productId from the user’s lifetime plan

xabg2:
In this PR, we obtain the `product Id` of the user's active lifetime plan in order to later obtain the features available for that product.

With this new function, we will be able to obtain the features using: `getTierProductsByProductsId` for lifetime users.


sg-gs:
The idea was to relate the tier with the user through a connection in-between so the API does not have to suffer Stripe's latency @xabg2. Otherwise lifetimes will create a big lag on the payments-server through long-lived requests to Stripe. Therefore, I recommend to create an in-between collection called `users_tiers` so you do not have to ask Stripe about it. 

The flow would be: 
- User buys a plan -> the webhook creates a users_tiers document relating user.id with tiers.id
- User refunds or cancels the plan -> the webhook deletes the users_tiers document by user.id
- User upgrades/downgrades a plan -> the webhook updates the users_tiers document by user.id

PR #168, opened 2025-02-19
[_]: feat/handling user-tier relationship

xabg2:
This PR implements the management of the user-tier relationship from the tiers service. Now, we control all the insertions/updates/deletions we throw an error indicating that it has gone wrong. 

This new implementation will allow us to handle cases where a subscription payment has been completed, the subscription has been updated or cancelled. 

In the other hand, this PR also introduces two new functions that enable the retrieval of `Tier` products based on a `user Id` instead of relying solely on the `product Id`. This enhancement allows for a more streamlined and user-centric approach when querying tier-related information, improving flexibility and maintainability.


tests/src/services/tiers.service.test.ts
```diff
@@ -74,7 +78,120 @@ describe('TiersService tests', () => {
...
+    });
+
+    it('When the requested tier exists, then it returns the tier object', async () => {
+      const tier = newTier();
+
+      jest.spyOn(tiersRepository, 'findByTierId').mockResolvedValue(tier);
+
+      const result = await tiersService.getTierProductsByTierId(tier.id);
+
+      expect(result).toEqual(tier);
```
sg-gs, line 192: use `toStrictEqual`, especially if you are matching objects

sg-gs (approved):

PR #169, opened 2025-02-20
[PB-3913/PB-3914]: feat/handle users tiers relationship

xabg2:
This PR adds a handler to inserts or updates the user-tier relationship and apply the tier to the user.
First, we check if the product is not mapped in the level collection. If this is the case, then we do nothing. But if the product is mapped, we check:
- The user-level relationship exists, then the relationship is updated.
- The user-level relationship does not exist, then the relationship is created.


src/services/tiers.service.ts
```diff
@@ -148,14 +163,20 @@ export class TiersService {
...
+  async applyDriveFeatures(
+    userWithEmail: User & { email: string },
+    customer: Stripe.Customer,
+    amountOfSeats: Stripe.InvoiceLineItem['quantity'],
+    tier: Tier,
+  ): Promise<void> {
     const features = tier.featuresPerService[Service.Drive];
 
     if (features.workspaces.enabled) {
+      if (!amountOfSeats) return;
```
sg-gs, line 175: if features.workspaces.enabled = true, `amountOfSeats` would be set always, therefore this check is not needed.

src/webhooks/utils/handleUserFeatures.ts
```diff
@@ -0,0 +1,50 @@
...
+    const [, latestInvoice] = userInvoices;
+
+    if (latestInvoice) {
+      let oldTierId;
+      const oldProductId = latestInvoice?.product as string;
+      const existingTier = existingTiersForUser.find((existingUserTier) => existingUserTier.productId === oldProductId);
+
+      if (!existingTier) return;
+
+      oldTierId = existingTier.id;
```
sg-gs, line 39: Either make this a const or remove the variable as declaring it as let 4 lines above due to no reason is obfuscating the code

tests/src/webhooks/utils/handleUserFeatures.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    expect(spyApplyTier).toHaveBeenCalledWith(
+      mockedUser,
+      mockedCustomer,
+      mockedPurchasedItem,
+      (mockedPurchasedItem.price?.product as Stripe.Product).id,
+    );
+    expect(spyUpdate).not.toHaveBeenCalled();
+  });
+
+  it('when the user has existing tiers and the second invoice has a product that is not mapped, then it should stop', async () => {
```
sg-gs, line 105: Be more specific about what means "stopping". I understand that it means that is does not apply any tier, hence, the question is: how are you going to handle a case where a user does not have a tier but somewhat this handler has been called? That seems like a non-expected path that should be errored so Stripe reports a failed webhook on its UI.

sg-gs (changes requested):

src/services/tiers.service.ts
xabg2: I have renamed the `amountOfSeats` parameter to `subscriptionSeats` for clarity. The idea is that this parameter represents the `amount` of subscription the user is paying for, which normally corresponds to the number of seats in the workspace plan - minimum of 3 seats per subscription. I have also added another check for seats less than the minimum.

sg-gs (approved):

PR #170, opened 2025-02-21
[PB-3915]: feat/remove tier for canceled subs and refunded lifetimes

xabg2:
This PR introduces a handler to properly manage user-tier relationships upon subscription cancellations or lifetime purchase refunds. It ensures that users are downgraded correctly while preventing unintended tier changes for lifetime purchases.

- **Lifetime Purchase Validation:**
  - If the user has purchased a lifetime plan and subsequently cancels an active subscription, we do not downgrade the storage or user tier.
  - This prevents removing storage benefits that should remain active.
  
- **Handling Subscription Cancellations & Refunds:**
  - If the lifetime condition does not apply:
    - Clear the cached subscription to ensure consistency.
    - Set lifetime to false (relevant for refunded lifetime purchases, and harmless otherwise).
    - Reset the user’s allocated storage to the free-tier limit.
    - Remove the user-tier relationship to detach any premium benefits.
    - Downgrade the user tier to the free plan.


sg-gs:
Nice description @xabg2

src/services/tiers.service.ts
```diff
@@ -177,6 +205,23 @@ export class TiersService {
...
+
+    if (features.workspaces.enabled) {
+      await this.usersService.destroyWorkspace(userUuid);
+      return;
+    }
+
+    try {
+      await updateUserTier(userUuid, FREE_INDIVIDUAL_TIER, this.config);
+    } catch (err) {
+      console.log(`[TIER/SUB_CANCELED] Error while updating user tier: uuid: ${userUuid} `);
```
sg-gs, line 219: The error is not being printed, if that happens, the log is useless

tests/src/services/tiers.service.test.ts
```diff
@@ -446,4 +581,31 @@ describe('TiersService tests', () => {
       expect(enableVPNTierSpy).not.toHaveBeenCalled();
     });
   });
+
+  describe('Remove VPN access based on user tier', () => {
+    it('When VPN is enabled in the canceled tier user, then a request to remove the VPN access for the user is sent', async () => {
```
sg-gs, line 586: This is a bit confusing. I assume you wanted to say: 

'When VPN was enabled on the cancelled tier, then the request to disable/remove it is sent'

sg-gs (changes requested):

sg-gs (dismissed):

PR #171, opened 2025-02-21
[_]: feat/subscriptions-with-trial-agreements

sg-gs:
### **What**

The capability of creating subscriptions with trials. Currently, this option is not using the database to make it dynamic and allow more trial reasons as we are not still sure about how this is going to be from the business standpoint, thus, we are trying not to create technical debt by coupling it to the current create subscription process

### **How**

A new endpoint and a new use case are added. We allow this feature by sending the proper trial code, which is an optional step to create a subscription on a theoretically integrated checkout as long as the code is correct. Now it only works for the incoming agreement we have with PcComponentes. 

We add a metadata field called 'why-trial' to identify the subscriptions affected by this trial once it ends in order to apply further subscription transformations.

⚠️  A new environment variable called `PC_CLOUD_TRIAL_CODE` is required!


sg-gs:
Environment variable added to production

PR #172, opened 2025-02-24
[_]: feat/extract payment succeed user flow

xabg2:
This PR refactors the handling of the invoice.payment_succeeded Stripe event for the old billing flow by extracting its logic into a separate function. The main goal is to facilitate future updates related to the new Tiers structure while keeping the codebase clean and modular.

**What:**
- Extracted the logic for processing successful invoice payments in the old flow into a dedicated function (handleOldInvoiceCompletedFlow).
- Ensured that the extracted function is fully reusable, allowing for easy implementation of the new form for Tiers.
- Improved maintainability by making it simpler to remove the old logic when it’s no longer needed.
- Maintained the same behavior to avoid breaking existing functionality.

**Why:**
- Modularity & Scalability: Separating the logic makes it easier to manage future updates when integrating the new Tiers.
- Easier Rollback & Maintenance: This approach ensures that we can easily switch between the old and new flows, reducing technical debt.
- Improved Code Readability: Isolating logic into a function improves clarity and testability.


src/webhooks/utils/handleOldInvoiceCompletedFlow.ts
```diff
@@ -0,0 +1,81 @@
...
+        throw err;
+      }
+
+      log.info(
+        `USER WITH CUSTOMER ID: ${customer.id} - UUID: ${userUuid} - EMAIL: ${
+          customer.email
+        } DOES NOT HAVE ANY WORKSPACE TO UPDATE, CREATING A NEW ONE`,
+      );
+      await usersService.initializeWorkspace(userUuid, {
+        newStorageBytes: Number(maxSpaceBytes),
```
sg-gs, line 61: Shouldn't the newStorageBytes be `maxSpaceBytes` * `subscriptionSeats` or am I missing something ?

sg-gs (changes requested):

src/webhooks/utils/handleOldInvoiceCompletedFlow.ts
xabg2: I just copied/pasted the logic we had in the handler. I guess this operation will be done in the drive-server side.

sg-gs (approved):

PR #173, opened 2025-02-25
[_]: feat/implement user drive updates

xabg2:
This PR integrates the previously extracted function for handling user subscriptions and workspace storage updates into the `invoice.payment_succeeded` webhook event. This ensures that user data is correctly updated when a payment is successfully processed, whether the user is subscribed to an individual plan or a business workspace.

**What:**
- Integrated the extracted function into the `invoice.payment_succeeded` webhook to handle both user subscriptions and workspace storage updates upon successful payment.
- Ensured that new users are properly initialized, while existing users have their subscriptions and storage correctly updated.
- Improved code organization and maintainability by centralizing user billing updates within the correct webhook event.

**Why:**
- Maintains modularity: Since the function was extracted in a previous PR, this change ensures that the billing logic is placed where it belongs without duplicating code.
- This will allow to us to mantain both old and new users, because the new users will have different products and tiers.

### DIFF
--- a/tests/src/webhooks/handleInvoiceCompleted.test.ts
+++ b/tests/src/webhooks/handleInvoiceCompleted.test.ts
@@ -215,6 +215,45 @@ describe('Process when an invoice payment is completed', () => {
       expect(updateUserTier).toHaveBeenCalledTimes(1);
       expect(usersRepository.insertUser).toHaveBeenCalledTimes(1);
     });
+
+    it('When updating user executes successfully, it should be called once with correct parameters', async () => {
+      const mockedInvoice = getInvoice({ status: 'paid' });
+      const mockedCustomer = getCustomer();
+      const mockedUser = getUser();
+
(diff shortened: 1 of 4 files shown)


tests/src/webhooks/handleInvoiceCompleted.test.ts
```diff
@@ -375,4 +499,99 @@ describe('Process when an invoice payment is completed', () => {
...
+        mockedInvoice,
+        usersService,
+        paymentService,
+        getLogger(),
+        cacheService,
+        config,
+        objectStorageService,
+      );
+
+      expect(storedCouponSpy).toHaveBeenCalledTimes(1);
```
sg-gs, line 594: I am approving it but please do check **always** the parameters of the function you are testing are correct. Assuming that because the execution reached this point is enough is making half of the work to ensure a strict test, if parameters are wrong, a lot of bugs have room to get in here even having this test in-between so the test become useless

sg-gs (dismissed):

PR #175, opened 2025-02-25
[PB-3887/PB-3913/PB-3914]: feat/handle new tier products after successful payment

xabg2:
This PR implements the new subscription tier flow that will soon be introduced in the payment system.

# **What:**
- If a productId is found in the subscription, it means the user is purchasing a new plan, and all features available for that product will be applied.
- For users with an existing subscription, the legacy flow remains unchanged, allowing them to keep their current plan. However, if they decide to upgrade, they will be required to switch to one of the new products.

# **Why:**
- To introduce a new tier system while ensuring backward compatibility with existing subscriptions.
- To enforce a structured transition where users on legacy plans can continue using them but must migrate to the new system upon upgrading.
- To maintain a seamless experience for both new and existing users while applying all features correctly based on the selected subscription.

⚠️ This implementation will need a new environment variable: `VPN_URL`

### DIFF
--- a/src/core/users/MongoDBTiersRepository.ts
+++ b/src/core/users/MongoDBTiersRepository.ts
@@ -2,7 +2,7 @@ import { Collection, MongoClient, ObjectId, WithId } from 'mongodb';
 import { Tier } from './Tier';
 
 export interface TiersRepository {
-  findByProductId(productId: Tier['productId'], billingType?: Tier['billingType']): Promise<Tier | null>;
+  findByProductId(where: Partial<Tier>): Promise<Tier | null>;
   findByTierId(tierId: Tier['id']): Promise<Tier | null>;
 }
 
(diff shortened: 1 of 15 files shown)


src/core/users/MongoDBTiersRepository.ts
```diff
@@ -20,11 +20,14 @@ export class MongoDBTiersRepository implements TiersRepository {
     this.collection = mongo.db('payments').collection<Tier>('tiers');
   }
 
-  async findByProductId(productId: Tier['productId']): Promise<Tier | null> {
-    const tier = await this.collection.findOne({
-      productId,
-    });
+  async findByProductId(productId: Tier['productId'], billingType?: Tier['billingType']): Promise<Tier | null> {
```
sg-gs, line 23: Why not

`async findByProductId(productId: Tier['productId'], where: Partial<Tier>): Promise<Tier | null>`
?

sg-gs: So you do not have to do the magic trick inside the function (unknowns, undefined, etc)

sg-gs (changes requested):

src/core/users/MongoDBTiersRepository.ts
xabg2: Much cleaner. Updating it 🚀

sg-gs (approved):

PR #176, opened 2025-02-25
[PB-3915]: feat/delete user tier in cancel sub webhook event

xabg2:
# **What:**

This pull request introduces functionality to remove the association between users and their respective tiers within our local databases. It ensures that upon certain webhook events, users' subscriptions are either canceled or downgraded to the free plan, effectively managing their access levels.

# **Why:**
Implementing this feature is crucial for maintaining accurate user-tier relationships and ensuring that our system reflects the current subscription status of each user. By automating the cancellation or downgrading process through webhook events, we enhance operational efficiency and provide users with a seamless experience that aligns with their subscription choices.


src/webhooks/handleLifetimeRefunded.ts
```diff
@@ -4,31 +4,58 @@ import CacheService from '../services/cache.service';
...
+    if (!(error instanceof TierNotFoundError)) {
+      throw error;
+    }
+    await usersService.updateUser(customerId, { lifetime: false });
 
-  return storageService.changeStorage(uuid, FREE_PLAN_BYTES_SPACE);
+    try {
+      await updateUserTier(uuid, FREE_INDIVIDUAL_TIER, config);
+    } catch (err) {
+      log.error(`Error while updating user tier: uuid: ${uuid} `);
```
sg-gs, line 54: Log this message _together_ with the `err` object, as we receive thousands of logs, it these ones do not go together, there is an inherent risk of losing the trace by not making the error searchable nor relatable to the customer that is affected by it. 

For instance, imagine you are watching the logs: 
```
<SOME_TIMESTAMP>: Error while updating user tier: uuid: <some_uuid>
.... 2543 logs here
<SOME_TIMESTAMP>: Unhandled error 'Tier not found' at ....
```

How are you going to relate both by a log query? Although they are related, we cannot establish the relation, therefore the first log is not too helpful.

sg-gs (approved):

PR #180, opened 2025-03-03
[_]: feat/cancelling-subscription-lifetime-logs

xabg2:
This PR implements logs to track flows when cancelling a subscription/lifetime. With these logs, we can keep track of when a user cancels a plan.

### DIFF
--- a/src/webhooks/handleSubscriptionCanceled.ts
+++ b/src/webhooks/handleSubscriptionCanceled.ts
@@ -1,8 +1,8 @@
 import { UserType } from './../core/users/User';
 import { FastifyBaseLogger, FastifyLoggerInstance } from 'fastify';
-import { FREE_INDIVIDUAL_TIER, FREE_PLAN_BYTES_SPACE } from '../constants';
+import { FREE_PLAN_BYTES_SPACE } from '../constants';
 import CacheService from '../services/cache.service';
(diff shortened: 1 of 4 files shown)


src/webhooks/handleSubscriptionCanceled.ts
```diff
@@ -101,15 +105,20 @@ export default async function handleSubscriptionCanceled(
...
     try {
       await updateUserTier(uuid, FREE_INDIVIDUAL_TIER, config);
     } catch (err) {
-      log.error(`[TIER/SUB_CANCELED] Error while updating user tier: uuid: ${uuid} `);
-      log.error(err);
+      const error = err as Error;
+      log.error(
+        `[SUB CANCEL/ERROR]: Error while updating user tier: uuid: ${uuid}. [ERROR STACK]: ${error.stack ?? error.message} `,
+      );
+      throw err;
```
sg-gs, line 121: Remove this throw, this is changing the behaviour of the payments service completely. The policy is to change user's conditions first and then the tier things after. (pay attention to the changeStorage call behind)

sg-gs (changes requested):

sg-gs:
If the PR is about adding logs, I do not understand the throw added there, that is a behaviour change @xabg2

PR #184, opened 2025-03-10
[PB-4041/PB-4043]: feat/stack multiple lifetime storage

xabg2:
## **What:**
This PR introduces the `handleStackLifetimeStorage` function. It allows to stack space when the user has a lifetime plan and buys another one. In this way, we can also control that the user keeps the highest tier, so that it only affects the drive space if the tier of the purchased lifetime is lower or the same as the one the user has right now.

## **Why:**
Previously, if the user purchased a lifetime plan, the space had to be modified by hand in the database directly. Since we have classified the products by tiers, we now need to update additional information besides the space.

For this reason we have created this new function that executes everything that needs to be done when a user, having a lifetime plan, buys another lifetime plan.


src/services/tiers.service.ts
```diff
@@ -154,10 +154,10 @@ export class TiersService {
   async applyTier(
     userWithEmail: { email: string; uuid: User['uuid'] },
     customer: Stripe.Customer,
-    lineItem: Stripe.InvoiceLineItem,
+    amountOfSeats: Stripe.InvoiceLineItem['quantity'],
     productId: string,
+    excludedServices?: Service[],
```
sg-gs, line 159: I understand why you need this one, but it would better, at least, to rename this variable to `alreadyEnabledServices`

sg-gs (dismissed):

sg-gs (approved):

PR #185, opened 2025-03-10
[PB-3908]: feat/apply stack of lifetime storage

xabg2:
## **What**
This PR handles stacking lifetimes cases - when the user purchases a plan (`invoice.payment-succeeded`) and when a lifetime is refunded (`charge.refund`).
## **Why**
We need to handle the stack of storage when the lifetime plan is paid. Also, we need to reset the user tier to free when the last lifetime is refunded.

### DIFF
--- a/src/webhooks/handleSubscriptionCanceled.ts
+++ b/src/webhooks/handleSubscriptionCanceled.ts
@@ -89,7 +89,7 @@ export default async function handleSubscriptionCanceled(
     log.error(`Error in handleSubscriptionCanceled after trying to clear ${customerId} subscription`);
   }
 
-  if (hasBoughtALifetime) {
+  if (hasBoughtALifetime && productType === UserType.Individual) {
     log.info(`User with uuid ${uuid} has a lifetime subscription. No need to downgrade the user.`);
     // This user has switched from a subscription to a lifetime, therefore we do not want to downgrade his space
     // The space should not be set to Free plan.
--- a/src/webhooks/handleSubscriptionCanceled.ts
+++ b/src/webhooks/handleSubscriptionCanceled.ts
@@ -90,6 +90,7 @@ export default async function handleSubscriptionCanceled(
   }
 
   if (hasBoughtALifetime) {
+    log.info(`User with uuid ${uuid} has a lifetime subscription. No need to downgrade the user.`);
     // This user has switched from a subscription to a lifetime, therefore we do not want to downgrade his space
     // The space should not be set to Free plan.
     return;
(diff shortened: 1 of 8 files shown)


src/webhooks/handleSubscriptionCanceled.ts
```diff
@@ -89,7 +89,8 @@ export default async function handleSubscriptionCanceled(
     log.error(`Error in handleSubscriptionCanceled after trying to clear ${customerId} subscription`);
   }
 
-  if (hasBoughtALifetime) {
+  if (hasBoughtALifetime && productType === UserType.Individual) {
```
sg-gs, line 92: Lifetimes are only for individuals, why is the `productType` comparison needed?

src/webhooks/utils/handleUserFeatures.ts
```diff
@@ -45,10 +46,31 @@ export const handleUserFeatures = async ({
 
   try {
     const existingUser = await usersService.findUserByUuid(user.uuid);
+    const isLifetimeStackTry = tier.billingType === 'lifetime' && existingUser.lifetime;
```
sg-gs, line 49: `isStackingLifetimes` could be a better naming

src/webhooks/utils/handleUserFeatures.ts
```diff
@@ -45,10 +46,31 @@ export const handleUserFeatures = async ({
...
 
     const isLifetimePlan = isBusinessPlan ? existingUser.lifetime : isLifetimeCurrentSub;
 
     const existingTiersForUser = await tiersService.getTiersProductsByUserId(existingUser.id);
+    const oldLifetimeTier = existingTiersForUser.find(
```
sg-gs, line 54: `currentTierIsLifetime` could be a better naming

src/webhooks/utils/handleUserFeatures.ts
```diff
@@ -45,10 +46,31 @@ export const handleUserFeatures = async ({
...
+      (existingUserTier) => existingUserTier.billingType === 'lifetime',
+    );
+
+    if (isLifetimeStackTry && oldLifetimeTier) {
+      logger.info(
+        `User with uuid ${user.uuid} has a lifetime and purchased the tier with id ${newTierId}. Updating user and tier...`,
+      );
+      handleStackLifetimeStorage({
```
sg-gs, line 62: Are you missing an `await` here?

sg-gs (changes requested):

sg-gs (approved):

PR #191, opened 2025-03-25
[_]: chore/migrate update storage endpoint

xabg2:
This PR migrates the changeStorage endpoint from the drive server to the WIP. 

The createOrUpdateUser function is also no longer used.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -56,4 +56,6 @@ if (undefinedMandatoryVariables.length) {
   throw new Error(`Some mandatory variables are undefined: ${undefinedMandatoryVariables.join(' - ')}.`);
 }
 
+export const isProduction = process.env.NODE_ENV === 'production';
+
 export default process.env as AppConfig;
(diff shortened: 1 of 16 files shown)


tests/src/services/storage.service.test.ts
```diff
@@ -1,19 +1,65 @@
...
+        expect.objectContaining({
+          headers: expect.objectContaining({
+            Authorization: expect.stringContaining('Bearer '),
+            'Content-Type': 'application/json',
+          }),
+        }),
+      );
+    });
+
+    it('When the updates the storage, then it should be done correctly and called with the necessary data', async () => {
```
sg-gs, line 51: How is this test checking that the required parameters are being enforced?

sg-gs (changes requested):

tests/src/services/storage.service.test.ts
xabg2: I have put the description of the tests backwards. Changing it.

sg-gs (approved):

PR #192, opened 2025-03-25
[PB-4137]: feat/get higher tier available for user

xabg2:
This PR implements a new endpoint to get the user's tier if he has one, depending on whether the individual tier or the business tier is requested.
- If the individual tier is requested, it is searched by the tier associated to the user. 
- If the business tier is requested, we search if the user has any active business subscription and if he belongs to a Workspace (as if he“belongs” to the active subscription of the owner of another workspace). We obtain the all tiers related to the workspaces he pertain, filter by the highest one and return it.

### DIFF
--- a/fastify-jwt.d.ts
+++ b/fastify-jwt.d.ts
@@ -6,10 +6,11 @@ declare module '@fastify/jwt' {
       payload: {
         email: string;
         uuid: string;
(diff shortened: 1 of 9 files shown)


xabg2:
Hey @sg-gs, could you take a look at this PR when you have some time? Thanks!

PR #193, opened 2025-03-27
[_]: feat/subscription-with-trial-for-pccomponentes

jaaaaavier:
This PR introduces several improvements related to the subscription system, focusing on the integration of a new trial type, pc-components-25. The changes include:

1. Update in config.ts:

- The environment variable PC_COMPONENTES_TRIAL_CODE has been added to the list of mandatory variables for use in workflows.

2. Modifications in payments.controller.ts:

- The trial token validation has been extended to allow the pc-cloud-25 and pc-components-25 trial types.

- Subscription creation has been updated to accept these new trial types during the registration process.

3. Update in payment.service.ts:

- pc-components-25 has been added as a valid trial type within the subscription logic and the maps related to free trial periods.

4. New tests in payment.service.test.ts:

- A new test case has been added to validate subscription creation with the pc-components-25 trial option, ensuring the correct calculation of the trial end date

### DIFF
--- a/src/controller/payments.controller.ts
+++ b/src/controller/payments.controller.ts
@@ -19,6 +19,7 @@ import {
   UserAlreadyExistsError,
   CustomerNotFoundError,
   InvalidTaxIdError,
+  Reason,
 } from '../services/payment.service';
 import { User, UserSubscription, UserType } from '../core/users/User';
 import CacheService from '../services/cache.service';
@@ -519,12 +520,14 @@ export default function (
         if (!trialToken) {
           return res.status(403).send('Invalid trial token');
         }
-
+        const validTrialValues: Reason['name'][] = ['prevent-cancellation', 'pc-cloud-25', 'pc-componentes-25'];
+       
         try {
           const payload = jwt.verify(trialToken, config.JWT_SECRET) as { trial?: string };
-          if (!payload.trial ||( payload.trial !== 'pc-cloud-25' && payload.trial !== 'pc-componenes-25')) {
+          if (!payload.trial || !validTrialValues.includes(payload.trial as Reason['name'])) {
             throw new Error('Invalid trial token');
           }
+          
         } catch {
           return res.status(403).send('Invalid trial token');
         }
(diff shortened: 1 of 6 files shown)


src/controller/payments.controller.ts
```diff
@@ -520,27 +520,29 @@ export default function (
...
+          };
           const subscriptionSetup = await paymentService.createSubscriptionWithTrial(
             {
               customerId,
               priceId,
               currency,
             },
             {
-              name: 'pc-cloud-25',
+            name: metadata.name as 'pc-cloud-25' | 'pc-componentes-25' | 'prevent-cancellation',
```
sg-gs, line 541: Please keep this value as it was: pc-cloud-25, as this will be used for the webhook after 6 months. Having 3 different metadata values is just adding complexity. It's okay to add a new acceptable trial value on the token that allows the customer to get those 6 months with a different token, but do it just for the controller check, leave the rest of the field the same.

sg-gs (changes requested):

src/controller/payments.controller.ts
jaaaaavier: Done!

src/services/payment.service.ts
```diff
@@ -67,7 +67,7 @@ export interface PaymentIntent {
 }
 
 export type Reason = {
-  name: 'prevent-cancellation' | 'pc-cloud-25';
+  name: 'prevent-cancellation' | 'pc-cloud-25' |'pc-componentes-25';
```
sg-gs, line 70: This is no longer needed if you removed the reason to use the same path

src/services/payment.service.ts
```diff
@@ -83,6 +83,7 @@ const additionalPaymentTypesForOneTime: Record<string, Stripe.Checkout.SessionCr
 const reasonFreeMonthsMap: Record<Reason['name'], number> = {
   'prevent-cancellation': 3,
   'pc-cloud-25': 6,
+  'pc-componentes-25': 6,
```
sg-gs, line 86: Same here

src/controller/payments.controller.ts
```diff
@@ -520,27 +520,29 @@ export default function (
...
 
+        const validTrials = ['pc-cloud-25', 'pc-componentes-25'];
+
         try {
           const payload = jwt.verify(trialToken, config.JWT_SECRET) as { trial?: string };
-          if (!payload.trial || payload.trial !== 'pc-cloud-25') {
+          if (!payload.trial || !validTrials.includes(payload.trial)) {
             throw new Error('Invalid trial token');
           }
-        } catch {
```
sg-gs, line 528: Why did you remove the 403? This could useful for the mobile app that includes the same offer

sg-gs (changes requested):

src/controller/payments.controller.ts
```diff
@@ -520,27 +520,29 @@ export default function (
...
-          return res.status(403).send('Invalid trial token');
-        }
-
-        try {
+          const metadata = {
```
xabg2, line 530: I do not see this metadata object being used. Remove it if it is the case.

src/controller/payments.controller.ts
```diff
@@ -520,27 +520,29 @@ export default function (
           return res.status(403).send('Invalid trial token');
         }
 
+        const validTrials = ['pc-cloud-25', 'pc-componentes-25'];
```
xabg2, line 523: Remove `pc-components-25` from here if it will no longer be used.

xabg2 (changes requested):

PR #208, opened 2025-05-07
[_]: feat/create subscription

xabg2:
This PR implements a new endpoint that creates a subscription (`POST /checkout/subscription`) in the checkout controller and also includes its tests.

### DIFF
--- a/src/controller/checkout.controller.ts
+++ b/src/controller/checkout.controller.ts
@@ -1,13 +1,13 @@
 import { FastifyInstance } from 'fastify';
+import Stripe from 'stripe';
 import jwt from 'jsonwebtoken';
 import fastifyJwt from '@fastify/jwt';
 import fastifyRateLimit from '@fastify/rate-limit';
 
-import config from '../config';
 import { UsersService } from '../services/users.service';
 import { PaymentService } from '../services/payment.service';
-import { UnauthorizedError } from '../errors/Errors';
-import Stripe from 'stripe';
+import { BadRequestError, ForbiddenError, UnauthorizedError } from '../errors/Errors';
+import config from '../config';
 
 function signUserToken(customerId: string) {
   return jwt.sign({ customerId }, config.JWT_SECRET);
@@ -21,7 +21,7 @@ export default function (usersService: UsersService, paymentsService: PaymentSer
       timeWindow: '1 minute',
     });
 
-    fastify.addHook('onRequest', async (request, reply) => {
+    fastify.addHook('onRequest', async (request, _) => {
       try {
         await request.jwtVerify();
       } catch (err) {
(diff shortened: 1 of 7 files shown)


src/controller/checkout.controller.ts
```diff
@@ -21,7 +21,7 @@ export default function (usersService: UsersService, paymentsService: PaymentSer
       timeWindow: '1 minute',
     });
 
-    fastify.addHook('onRequest', async (request, reply) => {
+    fastify.addHook('onRequest', async (request, _) => {
```
sg-gs, line 24: If you are not using the `reply`, just do not declare it on the function (aka remove the '_')

src/services/payment.service.ts
```diff
@@ -1035,6 +1036,50 @@ export class PaymentService {
...
+   * @param currency - The currency of the requested price
+   * @returns - The selected price if it exists and it is active
+   */
+  async getPriceById(priceId: string, currency = 'eur') {
+    const availablePrices = await this.getPricesRaw(currency);
+
+    const selectedPrice = availablePrices.find((price) => price.id === priceId && price.active);
+
+    if (!selectedPrice) {
+      throw new NotFoundError('The requested price does not exists');
```
sg-gs, line 1051: does not **exist**

tests/src/controller/checkout.controller.test.ts
```diff
@@ -116,4 +123,96 @@ describe('Checkout controller', () => {
...
+          },
+          headers: {
+            authorization: `Bearer ${authToken}`,
+          },
+        });
+
+        expect(response.statusCode).toBe(400);
+      });
+
+      it('When the provided user token is not valid, then an error indicating so is thrown', async () => {
```
sg-gs, line 196: This test made my think about the diff between not a valid token and a non-matching customer Id (`if (customerId !== tokenCustomerId)` ....). You are testing the second, what about the first? Also, you may reconsider the legend of this test as the token is valid in this case, just contains a different customer id (either an impersonation attempt or a code logic error in the frontend)

sg-gs (changes requested):

tests/src/controller/checkout.controller.test.ts
```diff
@@ -116,4 +123,139 @@ describe('Checkout controller', () => {
...
+            token: invalidUserToken,
+          },
+          headers: {
+            authorization: `Bearer ${authToken}`,
+          },
+        });
+
+        const responseBody = response.json();
+
+        expect(response.statusCode).toBe(500);
```
sg-gs, line 235: This should be a 401, FYI, as it is an authentication issue, but as that could be problematic right now for some frontends, use at least a 403.

sg-gs (dismissed):
I approve, but review the comment I set please

PR #231, opened 2025-05-29
[PB-4396]: feat/add event to create subscription and initialize object-storage 

xabg2:
## What
Add handling for Stripe’s `payment_intent.amount_capturable_updated` webhook that fires when the €1 card-verification hold becomes capturable.

## Why
If we do nothing, Stripe will settle that €1 as an actual charge a few days later. Voiding the intent keeps onboarding friction-free and prevents surprise charges.

## How
- Listen for `payment_intent.amount_capturable_updated`.
- Immediately cancel the Payment Intent, voiding the €1 hold.
- Create the Object-Storage subscription.
- Initilize the customer’s Object-Storage.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,3 +1,4 @@
 export const HUNDRED_TB = 109951162777600;
 export const FREE_PLAN_BYTES_SPACE = 1 * 1024 * 1024 * 1024;
 export const FREE_INDIVIDUAL_TIER = 'free_individual_tier';
+export const VERIFICATION_CHARGE = 100;
--- a/src/controller/payments.controller.ts
+++ b/src/controller/payments.controller.ts
@@ -32,6 +32,7 @@ import { fetchUserStorage } from '../utils/fetchUserStorage';
 import { TierNotFoundError, TiersService } from '../services/tiers.service';
 import { CustomerSyncService } from '../services/customerSync.service';
 import { ForbiddenError } from '../errors/Errors';
+import { VERIFICATION_CHARGE } from '../constants';
 
 type AllowedMethods = 'GET' | 'POST';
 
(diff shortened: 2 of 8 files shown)


xabg2:
@sg-gs **regarding the reliability of the event:**
The `payment_intent.amount_capturable_updated` event is reliable in our case because the Payment Intent is created using [manual capture mode](https://github.com/internxt/payments-server/blob/master/src/controller/payments.controller.ts#L517). This guarantees that Stripe emits the event only when the funds are confirmed and ready to be captured, which allows us to act at the exact moment we know the payment method is valid.

In addition, when the user attempts to confirm the Payment Intent, Stripe performs the fund verification for us, returning appropriate errors if there are insufficient funds or other issues. We simply wait for the `amount_capturable_updated` event to proceed, which signals that the required funds (in this case, €1) are available.

As an additional safeguard, we could also implement a check to ensure that the captured amount is exactly €1, validating that the authorization amount matches the expected verification value.

Further documentation about the event is available [here](https://docs.stripe.com/api/events/types#event_types-payment_intent.amount_capturable_updated).

**On the second point you mentioned:**
The user can indeed attempt the process multiple times. The /activation endpoint on the website only prevents activation if the Object Storage account is already created, so it does not block repeated attempts. Since the user is not creating a subscription but only triggering a €1 verification — and this Payment Intent is cancelled before the charge is finalized — the process can be retried as often as necessary without any actual charges.

That said, one point I may have overlooked is the case where the user already has an active subscription, but the Object Storage account creation fails. This could lead to a scenario where the user is subscribed but lacks access to the service. I will address this edge case accordingly.

xabg2:
This PR is ready for review @sg-gs

sg-gs (approved):

PR #232, opened 2025-06-05
[PB-4287]: feat/determine lifetime conditions

xabg2:
## What
This PR handles the logic for when a user purchases a lifetime plan. There are three possible user states to consider:
- The user is currently on a free plan
- The user has an active subscription
- The user already has a lifetime plan

## Why
Each case requires different handling:
- If the user is free, we simply apply the lifetime plan and allocate the corresponding storage/features.
- If the user has an active subscription, we first cancel the subscription before applying the lifetime plan.
- If the user already has a lifetime plan, we stack the storage by calculating the total allocated space based on valid previous lifetime purchases.

## How
We determine the user state and act accordingly:
- For free users and users with a subscription, we follow a seamless flow. If a subscription exists, it’s cancelled before applying the lifetime plan.
- For existing lifetime users, we retrieve all invoices associated with the customer, filter those related to lifetime plans, and ensure they have been paid and not refunded.
- Based on the number of valid lifetime invoices, we calculate and increase the user’s maxSpaceBytes accordingly.

### DIFF
--- a/tests/src/core/users/DetermineLifetimeConditions.test.ts
+++ b/tests/src/core/users/DetermineLifetimeConditions.test.ts
@@ -15,7 +15,7 @@ import testFactory from '../../utils/factory';
 import config from '../../../../src/config';
 import Stripe from 'stripe';
 import { DetermineLifetimeConditions } from '../../../../src/core/users/DetermineLifetimeConditions';
-import { getUser, getSubscription, newTier, getInvoices, getCustomer, getCharge } from '../../fixtures';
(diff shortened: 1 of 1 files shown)


xabg2:
This PR has been split into two smaller and more focused pull requests:
- [Handle stack lifetime storage](https://github.com/internxt/payments-server/pull/233).
- [Determine lifetime conditions](https://github.com/internxt/payments-server/pull/234) for user.

As a result, this PR is no longer needed and can be safely closed.

PR #233, opened 2025-06-09
[PB-4287]: feat/handle stack lifetime storage

xabg2:
This PR implements all the required functions to support lifetime plan stacking logic. It includes:
- Core logic to calculate and apply stacked lifetime plans
- Test coverage

### DIFF
--- a/tests/src/core/users/DetermineLifetimeConditions.test.ts
+++ b/tests/src/core/users/DetermineLifetimeConditions.test.ts
@@ -1,0 +1,214 @@
+import axios from 'axios';
+import { CouponsRepository } from '../../../../src/core/coupons/CouponsRepository';
+import { UsersCouponsRepository } from '../../../../src/core/coupons/UsersCouponsRepository';
+import { DisplayBillingRepository } from '../../../../src/core/users/MongoDBDisplayBillingRepository';
+import { TiersRepository } from '../../../../src/core/users/MongoDBTiersRepository';
+import { UsersTiersRepository } from '../../../../src/core/users/MongoDBUsersTiersRepository';
+import { ProductsRepository } from '../../../../src/core/users/ProductsRepository';
+import { UsersRepository } from '../../../../src/core/users/UsersRepository';
+import { Bit2MeService } from '../../../../src/services/bit2me.service';
+import { PaymentService } from '../../../../src/services/payment.service';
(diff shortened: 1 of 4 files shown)


tests/src/core/users/DetermineLifetimeConditions.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      const mockedTier = newTier({ billingType: 'lifetime' });
+
+      jest.spyOn(paymentService, 'getCustomer').mockResolvedValue(customer as Stripe.Response<Stripe.Customer>);
+      jest.spyOn(paymentService, 'getCustomersByEmail').mockResolvedValue([customer]);
+      jest.spyOn(paymentService, 'getInvoicesFromUser').mockResolvedValue([invoice]);
+      jest.spyOn(determineLifetimeConditions as any, 'getPaidInvoices').mockResolvedValue([invoice]);
+      jest.spyOn(tiersService, 'getTiersProductsByUserId').mockResolvedValue([mockedTier]);
+      jest.spyOn(determineLifetimeConditions as any, 'getHigherTier').mockResolvedValue(mockedTier);
+
+      const result = await (determineLifetimeConditions as any).handleStackingLifetime(user);
```
sg-gs, line 109: Although using `as any` is discouraged, if due to the setup is currently needed, do it in the first line, assigning it to a variable and then use that variable on the rest of the lines

tests/src/core/users/DetermineLifetimeConditions.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      invoice.paid_out_of_band = true;
+      invoice.lines.data[0].price!.metadata!.planType = 'one_time';
+      invoice.metadata = {};
+
+      const result = await (determineLifetimeConditions as any).getPaidInvoices(customer, [invoice]);
+
+      expect(result).toEqual([invoice]);
+    });
+
+    it('When the invoice is paid and it has not been refunded nor disputed, then the invoice is returned', async () => {
```
sg-gs, line 140: This case may be providing a false positive as the if will check one of those two status before and will not check the other as the code base uses an AND. Use one test case for refunded and other for disputed

sg-gs (changes requested):

sg-gs (approved):

PR #238, opened 2025-06-23
[_]: feature/expose redeemed coupons endpoint

xabg2:
This PR adds a new endpoint to retrieve all promotional codes associated with a user, based on the coupons we've previously tracked.
This allows the client to handle these codes when needed, improving the flexibility of features like exclusive theme activation for users who subscribed using specific coupons.

### DIFF
--- a/src/services/cache.service.ts
+++ b/src/services/cache.service.ts
@@ -3,6 +3,7 @@ import Redis from 'ioredis';
 import { type AppConfig } from '../config';
 
 const SUBSCRIPTION_EXPIRATION_IN_SECONDS = 15 * 60;
+const USED_COUPON_CODES_EXPIRATION_IN_SECONDS = 6 * 60 * 60;
 
 export default class CacheService {
   private readonly redis: Redis;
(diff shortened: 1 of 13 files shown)


src/services/cache.service.ts
```diff
@@ -30,11 +35,17 @@ export default class CacheService {
...
-    userType: UserType,
-    subscription: UserSubscription,
-  ): Promise<void> {
+  async getUsedUserPromoCodes(customerId: string): Promise<string[] | null> {
+    const cachedUsedPromoCodesByUser = await this.redis.get(this.buildUsedPromoCodesKey(customerId));
+
+    if (!cachedUsedPromoCodesByUser) {
+      return null;
+    } else {
+      return JSON.parse(cachedUsedPromoCodesByUser) as string[];
```
sg-gs, line 44: If it is not an object, why `JSON.parse`? And if it is, why `string[]`?

sg-gs (dismissed):

src/services/cache.service.ts
xabg2: Redis only accepts values of type `string`, `Buffer`, or `number` when using the set command (`this.redis.set` in this case).
Therefore, if we want to store an array in Redis, we need to first convert it into a JSON string using `JSON.stringify`,
and then retrieve and parse it using `JSON.parse` to convert it back into an array.

We can safely remove the unnecessary `as string[]` type assertion, but we do need to use `JSON.parse`, otherwise the returned value will remain a plain string.

sg-gs: Yeah, I understood that, but the `string[]` type is incorrect then, that is an object, not `string[]`, type it accordingly

PR #252, opened 2025-07-23
[PB-4719]: feat/apply tier if needed when redeeming license codes

xabg2:
This PR implements another way to redeem codes. If the code is associated with a price that corresponds to the level products, their characteristics will be added. Otherwise, we only change the storage.

### DIFF
--- a/tests/src/services/licenseCodes.service.test.ts
+++ b/tests/src/services/licenseCodes.service.test.ts
@@ -1,0 +1,321 @@
+import axios from 'axios';
+import Stripe from 'stripe';
+import { Bit2MeService } from '../../../src/services/bit2me.service';
+import {
+  InvalidLicenseCodeError,
+  LicenseCodeAlreadyAppliedError,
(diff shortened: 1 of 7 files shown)


tests/src/services/licenseCodes.service.test.ts
```diff
@@ -0,0 +1,321 @@
...
+        jest.spyOn(storageService, 'changeStorage').mockRejectedValue(unexpectedError);
+        const applyProductFeatures = licenseCodesService['applyProductFeatures'].bind(licenseCodesService);
+        await expect(applyProductFeatures(user, mockedCustomer, mockedLogger, maxSpaceBytes, null)).rejects.toThrow(
+          unexpectedError,
+        );
+      });
+    });
+
+    describe('Redeemed code with tier', () => {
+      test('When the user redeems a valid code that match with any tier, then all the features have been applied', async () => {
```
sg-gs, line 216: '....then the tier is applied'

PR #262, opened 2025-07-31
[PB-4406]: feat/enhance product tier features and improve tier selection logic

jzunigax2:
- Added 'cleaner' feature to the product features response.
- Refactored tier selection logic to prioritize business tiers for drive features.
- Introduced merging of non-drive features for better tier management.

### DIFF
--- a/src/controller/products.controller.ts
+++ b/src/controller/products.controller.ts
@@ -32,7 +32,10 @@ export default function (
 
     fastify.get(
       '/',
-      async (req, res): Promise<{ featuresPerService: { antivirus: boolean; backups: boolean } } | Error> => {
+      async (
+        req,
+        res,
+      ): Promise<{ featuresPerService: { antivirus: boolean; backups: boolean; cleaner: boolean } } | Error> => {
         let user: User;
         const userUuid = req.user.payload.uuid;
 
@@ -52,6 +55,7 @@ export default function (
               featuresPerService: {
                 antivirus: false,
                 backups: false,
+                cleaner: false,
               },
             });
           }
@@ -64,26 +68,14 @@ export default function (
       },
     );
 
-    fastify.get<{
-      Querystring: { subscriptionType?: 'individual' | 'business' };
-      schema: {
-        querystring: {
-          type: 'object';
-          properties: {
-            subscriptionType: { type: 'string'; enum: ['individual', 'business'] };
-          };
-        };
-      };
-    }>('/tier', async (req, rep): Promise<Tier | Error> => {
+    fastify.get('/tier', async (req, rep): Promise<Tier | Error> => {
       const userUuid = req.user.payload.uuid;
       const ownersId = req.user.payload.workspaces?.owners ?? [];
-      const subscriptionType = (req.query.subscriptionType as UserType) || UserType.Individual;
 
       try {
         const higherTier = await productsService.getApplicableTierForUser({
           userUuid,
           ownersId,
-          subscriptionType,
         });
 
         return rep.status(200).send(higherTier);
--- a/src/services/products.service.ts
+++ b/src/services/products.service.ts
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
 import { TierNotFoundError, TiersService } from './tiers.service';
 import { UserNotFoundError, UsersService } from './users.service';
 
+interface MergedTierFeatures {
+  mail: {
+    enabled: boolean;
+    addressesPerUser: number;
+    sourceTierId?: string;
+  };
+  meet: {
+    enabled: boolean;
+    paxPerCall: number;
+    sourceTierId?: string;
+  };
+  vpn: {
+    enabled: boolean;
+    featureId: string;
+    sourceTierId?: string;
+  };
+  antivirus: {
+    enabled: boolean;
+    sourceTierId?: string;
+  };
+  backups: {
+    enabled: boolean;
+    sourceTierId?: string;
+  };
+  drive: {
+    tier: Tier;
+  };
+}
+
+interface AllVpnFeatures {
+  vpnOptions: Array<{
+    featureId: string;
+    sourceTierId: string;
+    tierLabel: string;
+  }>;
+}
+
 export class ProductsService {
   constructor(
     private readonly tiersService: TiersService,
     private readonly usersService: UsersService,
   ) {}
 
-  async getApplicableTierForUser({
+  private selectBestDriveTier(businessTiers: Tier[], individualTiers: Tier[]): Tier {
+    if (businessTiers.length > 0) {
+      return businessTiers.reduce((best, current) => {
+        const bestStorage = best.featuresPerService[Service.Drive].workspaces.maxSpaceBytesPerSeat;
+        const currentStorage = current.featuresPerService[Service.Drive].workspaces.maxSpaceBytesPerSeat;
+
+        const isCurrentTierBetter = currentStorage > bestStorage;
+        return isCurrentTierBetter ? current : best;
+      }, businessTiers[0]);
+    }
+
+    if (individualTiers.length > 0) {
+      return individualTiers.reduce((best, current) => {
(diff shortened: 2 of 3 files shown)


xabg2:
Hey @jzunigax2 can you check the tests?

src/controller/products.controller.ts
```diff
@@ -32,7 +32,10 @@ export default function (
 
     fastify.get(
       '/',
-      async (req, res): Promise<{ featuresPerService: { antivirus: boolean; backups: boolean } } | Error> => {
+      async (
+        req,
+        res,
+      ): Promise<{ featuresPerService: { antivirus: boolean; backups: boolean; cleaner: boolean } } | Error> => {
```
xabg2, line 38: I believe this change is no longer necessary, as we will be using the other endpoint. It was temporary. You can ignore it.

Also, note that you must return `cleaner: <true | false>` in the `antivirusTier` object.

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+      return individualTiers.reduce((best, current) => {
+        const bestStorage = best.featuresPerService[Service.Drive].maxSpaceBytes;
+        const currentStorage = current.featuresPerService[Service.Drive].maxSpaceBytes;
+
+        const isCurrentTierBetter = currentStorage > bestStorage;
+        return isCurrentTierBetter ? current : best;
+      }, individualTiers[0]);
+    }
+
+    throw new TierNotFoundError('No drive tiers available');
```
xabg2, line 70: You can throw the Free Tier instead, so it's easier to handle on the client side — as we will use `<feature>.enabled`.

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+        paxPerCall: meetFeatures.paxPerCall,
+        sourceTierId: tier.id,
+      };
+    }
+  }
+
+  private mergeVpnFeatures(tier: Tier, mergedFeatures: MergedTierFeatures): void {
+    const vpnFeatures = tier.featuresPerService[Service.Vpn];
+    // We cannot determine which featureId provides access to more zones
+    // maybe add this info to the tier object?
```
xabg2, line 108: To better manage VPN access per tier, we could include an `availableZones` field within the VPN object or something like that. What do you think @sg-gs ?

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+          return ownerTiers.filter((tier) => tier.featuresPerService[Service.Drive].workspaces.enabled);
+        } catch (error) {
+          if (error instanceof UserNotFoundError) return [];
+          throw error;
+        }
+      });
+
+      const ownerTiersArrays = await Promise.all(ownerTierPromises);
+      const businessTiers = ownerTiersArrays.flat();
+      console.log('businessTiers', businessTiers);
```
xabg2, line 179: Ops, remove this console.log

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+    const uniqueTiers = availableTiers.filter((tier, index, self) => index === self.findIndex((t) => t.id === tier.id));
+
+    return uniqueTiers;
+  }
+
+  async getApplicableTierForUser({ userUuid, ownersId }: { userUuid: string; ownersId?: string[] }): Promise<Tier> {
+    const availableTiers = await this.collectAllAvailableTiers(userUuid, ownersId);
+
+    if (availableTiers.length === 0) {
+      throw new TierNotFoundError(`No tiers found for user uuid ${userUuid}`);
```
xabg2, line 192: Just use `NotFoundError`. It already triggers a 404 thanks to the global middleware we added (works like in NestJS).

References:
- [NotFoundError](https://github.com/internxt/payments-server/blob/master/src/errors/Errors.ts#L9).
- [Error handler](https://github.com/internxt/payments-server/blob/master/src/plugins/error-handler.ts)

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
-            }
-          });
-
-          const ownerTiersResolved = await Promise.all(ownerTierPromises);
-          const validOwnerTiers = ownerTiersResolved.filter((tier): tier is Tier => tier !== null);
-
-          tiers.push(...validOwnerTiers);
-        }
+    if (availableTiers.length === 0) {
+      throw new TierNotFoundError(`No tiers found for user uuid ${userUuid}`);
```
xabg2, line 217: Same here, use `NotFoundError` instead.

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+    return this.mergeFeatures(availableTiers);
+  }
 
-        return [...tiers].sort(
-          (a, b) =>
-            b.featuresPerService[Service.Drive].workspaces.maxSpaceBytesPerSeat -
-            a.featuresPerService[Service.Drive].workspaces.maxSpaceBytesPerSeat,
-        )[0];
-      }
+  async getFeatureSourceBreakdown({ userUuid, ownersId }: { userUuid: string; ownersId?: string[] }): Promise<{
```
xabg2, line 223: Is this function used?

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+      backups: string | null;
+      drive: string;
+    };
+  }> {
+    const availableTiers = await this.collectAllAvailableTiers(userUuid, ownersId);
 
-      default:
-        throw new TierNotFoundError(`Tier for ${userUuid} was not found while getting user Tier`);
+    if (availableTiers.length === 0) {
+      throw new TierNotFoundError(`No tiers found for user uuid ${userUuid}`);
```
xabg2, line 238: Same here, use `NotFoundError`

src/services/products.service.ts
```diff
@@ -3,66 +3,277 @@ import { UserType } from '../core/users/User';
...
+      .map((tier) => ({
+        featureId: tier.featuresPerService[Service.Vpn].featureId,
+        sourceTierId: tier.id,
+        tierLabel: tier.label,
+      }));
+
+    return { vpnOptions };
+  }
+
+  async hasVpnAccess({ userUuid, ownersId }: { userUuid: string; ownersId?: string[] }): Promise<boolean> {
```
xabg2, line 275: is this function used?

src/services/products.service.ts
jzunigax2: oops, my bad. deleted the two unused functions

src/services/products.service.ts
```diff
@@ -1,68 +1,221 @@
...
+
+  private mergeFeatures(tiers: Tier[]): MergedTierFeatures {
+    const mergedFeatures: MergedTierFeatures = {
+      mail: { enabled: false, addressesPerUser: 0 },
+      meet: { enabled: false, paxPerCall: 0 },
+      vpn: { enabled: false, featureId: '' },
+      antivirus: { enabled: false },
+      backups: { enabled: false },
+      cleaner: { enabled: false },
+      drive: { tier: tiers[0] },
```
xabg2, line 161: Why do you return the entire tier for drive? isn't it enough returning `tiers[0].featuresPerService[Service.Drive]`?

src/services/products.service.ts
```diff
@@ -1,68 +1,221 @@
...
+    const user = await this.usersService.findUserByUuid(userUuid);
+    if (user.lifetime) {
+      const lifetimeTier = availableTiers.find((tier) => tier.billingType === 'lifetime');
+      if (lifetimeTier) {
+        return lifetimeTier;
+      }
     }
+
+    const mergedFeatures = this.mergeFeatures(availableTiers);
+    return mergedFeatures.drive.tier;
```
xabg2, line 219: Not sure why we return the drive tier instead of the `mergedFeatures`.

PR #265, opened 2025-08-01
[_]: fix/return the correct payload when creating an invoice

xabg2:
This PR returns the correct payload, returning the QR URL, addressRequestUri, etc from crypto invoices.

### DIFF
--- a/tests/src/fixtures.ts
+++ b/tests/src/fixtures.ts
@@ -59,16 +59,16 @@ export const getCryptoInvoiceWebhook = (params?: Partial<Bit2MePaymentStatusCall
     id: randomDataGenerator.string({ length: 16 }),
     foreignId: `inv_${randomDataGenerator.string({ length: 16 })}`,
     cryptoAddress: {
-      currency: AllowedCurrencies['Bitcoin'],
+      currency: AllowedCryptoCurrencies['Bitcoin'],
       address: randomDataGenerator.hash({ length: 34 }),
     },
     currencySent: {
-      currency: AllowedCurrencies['Bitcoin'],
+      currency: AllowedCryptoCurrencies['Bitcoin'],
       amount: '0.01',
       remainingAmount: '0',
     },
     currencyReceived: {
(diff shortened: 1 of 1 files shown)


src/services/payment.service.ts
```diff
@@ -414,10 +415,12 @@ export class PaymentService {
     additionalInvoiceOptions?: Partial<Stripe.InvoiceCreateParams>;
   }): Promise<PaymentIntent> {
     let couponId: string | undefined = undefined;
+    const isCryptoCurrency = this.bit2MeService.isAllowedCurrency(currency);
+    const currencyForStripeInvoice = isCryptoCurrency ? 'eur' : currency;
```
sg-gs, line 419: Does the Bit2Me format coincide with the Stripe's one? I would add a layer between that, otherwise, Bit2Me changing the API currency format will break the process

tests/src/controller/checkout.controller.test.ts
```diff
@@ -331,6 +331,56 @@ describe('Checkout controller', () => {
       expect(responseBody).toStrictEqual(mockedPaymentIntent);
     });
 
+    it('when the user want to pay a one time plan using crypto currencies, then an invoice is created and the specific payload containing the QR Link is returned', async () => {
```
sg-gs, line 334: I miss a test to ensure that when the interval is not lifetime, the system handles it. We would not like to find someone paying a sub with crypto hahaha

sg-gs (changes requested):

src/services/payment.service.ts
xabg2: Good catch! I've verified the formats:

- **Stripe**: Uses lowercase ISO 4217 codes (eur, usd)
- **Bit2Me**: Uses uppercase format (EUR, USD, BTC, ETH)

Working on it! 🚀

xabg2:
This request to incorporate changes depends on [this other one](https://github.com/internxt/payments-server/pull/267), in which we created a currency adapter to work with the appropriate format for each service.

@sg-gs

PR #267, opened 2025-08-07
[PB-3669]: feature/currency adapter

xabg2:
This PR implements a currency format adapter to handle the differences between Stripe and Bit2Me APIs. Stripe uses lowercase ISO 4217 currency codes (e.g., "eur", "usd") while Bit2Me uses uppercase format (e.g., "EUR", "USD", "BTC", "ETH"). Without proper format handling, API calls can fail due to currency code mismatches. The solution introduces a `CurrencyAdapter` that normalizes currency codes to the appropriate format for each provider, validates currency support before making API calls, and centralizes all currency-related logic.


src/services/currencyAdapter.service.ts
```diff
@@ -0,0 +1,68 @@
...
+  Cardano = 'ADA',
+  BinanceCoin = 'BNB',
+}
+
+export enum AllowedFiatCurrencies {
+  Euro = 'EUR',
+  USDollar = 'USD',
+}
+
+export class CurrencyAdapter {
```
sg-gs, line 21: If everything is static, why don't just use functions?

tests/src/services/currencyAdapter.service.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+
+    test('When a valid cryptocurrency is provided, then it returns fallback fiat currency that is eur in lowercase format', () => {
+      const result = CurrencyAdapter.normalizeForStripe('BTC');
+
+      expect(result).toBe('eur');
+    });
+
+    test('When an unsupported currency is provided, then it throws an error', () => {
+      expect(() => CurrencyAdapter.normalizeForStripe('INVALID')).toThrow('Currency INVALID is not supported');
```
sg-gs, line 28: If you can create a custom error and check the instaceof it, better, as that makes the test easier to maintain, otherwise, changing the error message will require a change here

tests/src/services/currencyAdapter.service.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    test('When currency has whitespace, then it normalizes correctly', () => {
+      const result = CurrencyAdapter.normalizeForStripe('  EUR  ');
+
+      expect(result).toBe('eur');
+    });
+  });
+
+  describe('Normalize currency for Bit2Me', () => {
+    test('When a fiat currency is provided, then an error indicating that Bit2Me does not support these currencies is thrown', () => {
+      expect(() => CurrencyAdapter.normalizeForBit2Me('EUR')).toThrow('Currency EUR is not supported by Bit2Me');
```
sg-gs, line 40: Same here

tests/src/services/currencyAdapter.service.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    const invalidCurrencies = ['INVALID', 'XYZ', '123', ''];
+
+    test.each(invalidCurrencies)('When invalid currency "%s" is provided, then it returns false', (currency) => {
+      expect(CurrencyAdapter.isValidCurrency(currency)).toBe(false);
+    });
+  });
+
+  describe('edge cases', () => {
+    test('When empty string is provided, then appropriate error is thrown', () => {
+      expect(() => CurrencyAdapter.normalizeForStripe('')).toThrow();
```
sg-gs, line 115: Same here. Also, having a lower or uppercase could fit here, but the controller exists to ensure the format is valid. Therefore, receiving empty currencies on a service means the controller is useless, I would reconsider if that responsibility of checking empty / with space strings should be part of the service

tests/src/services/currencyAdapter.service.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      expect(() => CurrencyAdapter.normalizeForBit2Me('')).toThrow();
+    });
+
+    test('When currency with special characters is provided, then it handles gracefully', () => {
+      expect(() => CurrencyAdapter.normalizeForStripe('BTC!')).toThrow();
```
sg-gs, line 120: Same here

sg-gs (changes requested):

tests/src/services/currencyAdapter.service.test.ts
xabg2: I can use the generic `BadRequestError` to check what error was thrown instead of the message.

tests/src/services/currencyAdapter.service.test.ts
xabg2: Same here.

src/services/currencyAdapter.service.ts
xabg2: Ah, understood. It's better to add this outside of services` 👍🏽

tests/src/services/currencyAdapter.service.test.ts
xabg2: I kept this extreme case to make sure that the currency is correct; otherwise, we generate an error indicating this.

PR #269, opened 2025-08-13
[_]: test/test services factory

xabg2:
This PR creates a service factory so that we can use them in all tests without having to create them every time we need them.

### DIFF
--- a/tests/src/core/users/DetermineLifetimeConditions.test.ts
+++ b/tests/src/core/users/DetermineLifetimeConditions.test.ts
@@ -1,79 +1,15 @@
-import axios from 'axios';
-import { CouponsRepository } from '../../../../src/core/coupons/CouponsRepository';
-import { UsersCouponsRepository } from '../../../../src/core/coupons/UsersCouponsRepository';
-import { DisplayBillingRepository } from '../../../../src/core/users/MongoDBDisplayBillingRepository';
-import { TiersRepository } from '../../../../src/core/users/MongoDBTiersRepository';
-import { UsersTiersRepository } from '../../../../src/core/users/MongoDBUsersTiersRepository';
-import { ProductsRepository } from '../../../../src/core/users/ProductsRepository';
-import { UsersRepository } from '../../../../src/core/users/UsersRepository';
-import { Bit2MeService } from '../../../../src/services/bit2me.service';
-import { PaymentService } from '../../../../src/services/payment.service';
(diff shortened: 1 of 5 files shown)


tests/src/core/users/DetermineLifetimeConditions.test.ts
```diff
@@ -83,10 +19,10 @@ describe('Determining Lifetime conditions', () => {
       const tierNotFoundError = new TierNotFoundError('Old product was found');
       const mockedUser = getUser();
 
-      jest.spyOn(paymentService, 'getUserSubscription').mockResolvedValue({ type: 'free' });
-      jest.spyOn(tiersService, 'getTierProductsByProductsId').mockRejectedValue(tierNotFoundError);
+      jest.spyOn(services.paymentService, 'getUserSubscription').mockResolvedValue({ type: 'free' });
```
sg-gs, line 22: Consider destructuring the `services` object, reducing the LOCs changed and making it easier/shorter to read

sg-gs (changes requested):

tests/src/core/users/DetermineLifetimeConditions.test.ts
```diff
@@ -325,7 +262,7 @@ describe('Determining Lifetime conditions', () => {
 
       jest.spyOn(tiersService, 'getTierProductsByProductsId').mockResolvedValue(tierFromProduct);
 
-      const result = await determineConditions.getHigherTier([productId], null);
+      const result = await (determineLifetimeConditions as any).getHigherTier([productId], null);
```
sg-gs, line 265: Why is the any needed always if the return type of the factory is typed?

xabg2: Because they are private functions.

sg-gs: How are they private? They are a return of the factory

image

xabg2: I mean, `getHigherTier` is a private function inside `determineLifetimeConditions`

sg-gs (approved):

tests/src/core/users/DetermineLifetimeConditions.test.ts
xabg2: I tink I cannot spy `Class['functionKey']`. Maybe the `@ts-ignore` approach could do the work. I'm going to try this way.

PR #273, opened 2025-08-14
[PB-4735]: chore/bump stripe version to v16.12.0

xabg2:
This PR updates the Stripe SDK from v15.10.0 to  v16.12.0. We need to update the Stripe SDK to v18.X in order to use `Klarna` as an allowed payment method when creating invoices.


sg-gs (dismissed):
Which are the main changes introduced by Stripe to the new version vs the current one?

PR #274, opened 2025-08-14
[PB-4735]: chore/bump stripe version to v17.7.0

xabg2:
This PR updates the Stripe SDK from v16.12.0 to v17.7.0. We need to update the Stripe SDK to v18.X in order to use Klarna as an allowed payment method when creating invoices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "ioredis": "^5.4.1",
     "jsonwebtoken": "^9.0.0",
     "mongodb": "^6.10.0",
-    "stripe": "=16.12.0",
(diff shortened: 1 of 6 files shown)


sg-gs (dismissed):
Would you like to run a QA or something here? Just to ensure Stripe's integration is not affected. Tests are OK, but we do not have integration ones for these cases.

xabg2:
> Would you like to run a QA or something here? Just to ensure Stripe's integration is not affected. Tests are OK, but we do not have integration ones for these cases.

I did some tests in local but, do you have smth in mind to run a QA?

PR #287, opened 2025-08-27
[PB-4739]: refactor/remove useless event handler

xabg2:
This PR removes useless event handler for `invoice.payment_succeeded` since we already switched to the improved one.


sg-gs (approved):
omg

PR #291, opened 2025-08-28
[PB-4702]: feature/add klarna as accepted payment method type for invoices

xabg2:
This PR implements Klarna as a new accepted payment method, so the users can purchase one time plans (lifetimes) using that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "jsonwebtoken": "^9.0.0",
     "mongodb": "^6.10.0",
     "stripe": "=17.7.0",
+    "stripe-next-version": "npm:stripe@=18.5.0",
     "uuid": "^11.1.0"
   }
 }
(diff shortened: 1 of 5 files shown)


tests/src/services/payment.service.test.ts
```diff
@@ -250,36 +309,36 @@ describe('Payments Service tests', () => {
...
         jest
-          .spyOn(stripe.invoiceItems, 'create')
+          .spyOn(stripeNewVersion.invoiceItems, 'create')
           .mockResolvedValue(mockedInvoice.lines.data[0] as unknown as Stripe.Response<Stripe.InvoiceItem>);
 
-        jest.spyOn(stripe.invoices, 'update').mockImplementation();
+        jest.spyOn(stripeNewVersion.invoices, 'update').mockImplementation();
         jest
-          .spyOn(stripe.invoices, 'retrieve')
+          .spyOn(stripeNewVersion.invoices, 'retrieve')
```
xabg2, line 320: This spyOn does not work correctly; it throws an error indicating that the fake invoice ID does not exist. It is as if it were searching Stripe as such, instead of using the simulated value in spyOn.

sg-gs (dismissed):

PR #293, opened 2025-09-01
[_]: fix/insert user in DB after creating customer

xabg2:
This PR adds the insertion of the user into the database immediately after the customer is created. This change prevents the creation of multiple customers in case a payment fails, since previously the system could not verify whether the user already had a customer until the purchase was completed, at which point the user was added to the database.

This PR depends on the following:
- https://github.com/internxt/payments-server/pull/294

### DIFF
--- a/src/controller/checkout.controller.ts
+++ b/src/controller/checkout.controller.ts
@@ -91,6 +91,12 @@ export default function (usersService: UsersService, paymentsService: PaymentSer
               postal_code: postalCode,
             },
           });
+          await usersService.insertUser({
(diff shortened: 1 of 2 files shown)


xabg2:
> Let's add the recaptcha first to avoid opening the door for the db to be flooded with slightly different variations of the same email (as a way of spamming us)

Okay. I just added that because we have like 4 customers with the same email because they failed the payment (3D Secure failed, blocked cards, etc). So we have just one customer per user.

But yeah, let's add the captcha first.

PR #294, opened 2025-09-01
[PB-4876]: feature/captcha token when creating customer

xabg2:
This PR adds a captcha verification token when the user attempts to create a customer. 

⚠️ We will need the following variables in the .env of production:
- `RECAPTCHA_V3_ENDPOINT`
- `RECAPTCHA_V3_SCORE_THRESHOLD`
- `RECAPTCHA_V3`

### DIFF
--- a/tests/src/controller/checkout.controller.test.ts
+++ b/tests/src/controller/checkout.controller.test.ts
@@ -298,6 +298,33 @@ describe('Checkout controller', () => {
 
         expect(response.statusCode).toBe(403);
       });
+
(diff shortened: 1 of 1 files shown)


xabg2:
> Do you have the token configured on the frontend to be sent here? Otherwise this will break the payment process @xabg2

I am waiting your approval so I will deploy the new SDK version, add it to the frontend and open the PR, so we can deploy drive-web and payments at the same time.

sg-gs:
Environment variables added @xabg2

PR #296, opened 2025-09-02
[_]: fix/use user-tier relationship to return the available products

xabg2:
This PR uses the user-tier relationship as source of truth to return the available products for the user.

### DIFF
--- a/src/controller/products.controller.ts
+++ b/src/controller/products.controller.ts
@@ -64,7 +64,7 @@ export default function (
           const tier = {
             featuresPerService: {
               antivirus: mergedFeatures.featuresPerService.antivirus.enabled,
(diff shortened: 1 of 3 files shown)


src/controller/products.controller.ts
```diff
@@ -44,9 +44,31 @@ export default function (
 
           const isLifetimeUser = lifetime ?? false;
 
-          const antivirusTier = await tiersService.getProductsTier(customerId, isLifetimeUser);
+          const userSubscriptions = await paymentService.getActiveSubscriptions(customerId);
```
sg-gs, line 47: To speed up this check for the free users (the big part of the client base) I would check first against our db, if possible and just after, to Stripe, in that order

PR #304, opened 2025-09-19
[_]: feature/cache user tier

xabg2:
This PR implements Redis caching for user tier data with a 6-hour TTL to improve API performance. Since user subscription tiers rarely change, this reduces database queries.


sg-gs (approved):
LGTM. I am assuming that the InvoiceCompletedHandler cache cleanup is performed in any of this events (as it should be): cancelling / upgrading sub / refunds

PR #306, opened 2025-09-25
[_]: feature/merge both individual and business tiers

xabg2:
This PR merges all user tiers into one to return all available products from their individual and B2B plans to the user.


xabg2:
I think the test I've added is enough. Sonarcloud says we need to check the other case (if the product is activated in the individual plan instead of the B2B plan, for example). I would ignore Sonar this time.

PR #309, opened 2025-10-02
[PB-4944]: feat/pass drive tier id when updating drive storage

xabg2:
This PR modifies the Drive storage update flow to include the tierId when a user purchases or cancels a subscription.

Previously, only the storage space (`maxSpaceBytes`) was updated, but now Drive needs to know the associated tier to apply the correct policies and configurations.

### DIFF
--- a/src/services/licenseCodes.service.ts
+++ b/src/services/licenseCodes.service.ts
@@ -186,7 +186,7 @@ export class LicenseCodesService {
           await this.tiersService.insertTierToUser(userId, tierProduct.id);
         }
       } else {
-        await this.storageService.changeStorage(user.uuid, maxSpaceBytes);
+        await this.storageService.updateUserStorageAndTier(user.uuid, maxSpaceBytes, '');
       }
     } catch (error) {
       const err = error as Error;
--- a/src/services/licenseCodes.service.ts
+++ b/src/services/licenseCodes.service.ts
@@ -186,7 +186,12 @@ export class LicenseCodesService {
           await this.tiersService.insertTierToUser(userId, tierProduct.id);
         }
       } else {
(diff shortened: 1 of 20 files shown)


src/services/licenseCodes.service.ts
```diff
@@ -186,7 +186,7 @@ export class LicenseCodesService {
           await this.tiersService.insertTierToUser(userId, tierProduct.id);
         }
       } else {
-        await this.storageService.changeStorage(user.uuid, maxSpaceBytes);
+        await this.storageService.updateUserStorageAndTier(user.uuid, maxSpaceBytes, '');
```
sg-gs, line 189: Why an empty tier?

src/webhooks/events/invoices/InvoiceCompletedHandler.ts
```diff
@@ -272,7 +272,7 @@ export class InvoiceCompletedHandler {
    * @param maxSpaceBytes The new max space bytes
    */
   private handleOldProduct(userUuid: string, maxSpaceBytes: number): Promise<void> {
-    return this.storageService.changeStorage(userUuid, maxSpaceBytes);
+    return this.storageService.updateUserStorageAndTier(userUuid, maxSpaceBytes, '');
```
sg-gs, line 275: Why empty tier?

sg-gs (commented):
Should I add something to the production database @xabg2 ?

src/services/licenseCodes.service.ts
xabg2: Finally, I used the free tier for products that do not have any associated tier.

xabg2:
As we need to pass the `foreignTierId` when we create or update a workspace, I will pass this PR to draft while we work on the EP that belongs to Drive.

apsantiso (approved):

PR #310, opened 2025-10-03
[PB-4945]: feat/command to sync users with tiers in drive

xabg2:
This PR implements a new command in the CLI folder to be able to sync the user with their tier in Drive. 

By passing the business parameter, it will be sync the business subscriptions. By default it syncs the Individual ones (subscription/lifetime).

### DIFF
--- a/src/core/users/MongoDBUsersTiersRepository.ts
+++ b/src/core/users/MongoDBUsersTiersRepository.ts
@@ -9,6 +9,7 @@ export interface UserTier {
 }
 
 export interface UsersTiersRepository {
+  getUserTierMappings(isBusiness?: boolean): Promise<Array<{ userUuid: string; foreignTierId: string }>>;
   insertTierToUser(userId: User['id'], tierId: Tier['id']): Promise<void>;
   updateUserTier(userId: User['id'], oldTierId: Tier['id'], newTierId: Tier['id']): Promise<boolean>;
   deleteTierFromUser(userId: User['id'], tierId: Tier['id']): Promise<boolean>;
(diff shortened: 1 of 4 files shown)


apsantiso (dismissed):

src/services/users.service.ts
```diff
@@ -225,10 +225,15 @@ export class UsersService {
   }: {
     ownerId: string;
     tierId: string;
-    maxSpaceBytes: number;
-    seats: number;
+    maxSpaceBytes?: number;
+    seats?: number;
   }): Promise<void> {
     const jwt = signToken('5m', this.config.DRIVE_NEW_GATEWAY_SECRET);
+    let totalMaxSpaceBytes: number | undefined = undefined;
```
sg-gs, line 232: Why not `let totalMaxSpaceBytes = maxSpaceBytes` here? The param is also optional and if it set, it will go inside the if

src/core/users/MongoDBUsersTiersRepository.ts
```diff
@@ -30,6 +31,55 @@ export class MongoDBUsersTiersRepository implements UsersTiersRepository {
...
+          },
+        },
+        {
+          $unwind: '$tier',
+        },
+        {
+          $project: {
+            _id: 0,
+            userUuid: '$user.uuid',
+            foreignTierId: '$tier.foreignTierId',
```
sg-gs, line 74: How is this working if foreignTierId is inside tier.drive field?

sg-gs (changes requested):

src/services/users.service.ts
sg-gs: If it not being used, it should not be passed I assume, but this works also

sg-gs (approved):
The Coverage is on the floor, we need that check to pass first

xabg2:
> The Coverage is on the floor, we need that check to pass first

Okay, I will add coverage to the script

PR #351, opened 2026-02-11
[MKT-748]:feat/implement tracking services for cancelled subscriptions

jaaaaavier:
This task, which is linked to additional PRs in other repositories, aims to track users who cancel their plans and send that data to Klaviyo. While Klaviyo already integrates with Stripe, it lacks this specific functionality, which is why we created this solution. 

Changes:
- Created the klaviyo.service that sends the email of the user that cancels the subscription.
- Added the service on the handleSubscription

### Comment
Need to add two Environment variables


xabg2 (commented):
Add the tests for this new klaviyo service.

tests/src/services/klaviyo.service.test.ts
```diff
@@ -0,0 +1,102 @@
...
+describe('KlaviyoTrackingService', () => {
+  let service: KlaviyoTrackingService;
+  const mockApiKey = 'pk_test_12345';
+  const mockBaseUrl = 'https://a.klaviyo.com/api';
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    (config as any).KLAVIYO_API_KEY = mockApiKey;
+    (config as any).KLAVIYO_BASE_URL = mockBaseUrl;
+    service = new KlaviyoTrackingService();
```
xabg2: You can instantiate the service in the service-factory.

src/services/klaviyo.service.ts
```diff
@@ -0,0 +1,76 @@
...
+          Authorization: `Klaviyo-API-Key ${this.apiKey}`,
+          'Content-Type': 'application/json',
+          revision: '2024-10-15',
+        },
+      });
+
+      Logger.info(`[Klaviyo] ${eventName} tracked for ${email}`);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : 'Unknown error';
+      Logger.error(`[Klaviyo] ${eventName} failed for ${email}:`, message);
```
xabg2: Add the error inside the first string, I think it is not printed if it is outside. Or it works?

xabg2 (dismissed):

xabg2 (approved):

PR #364, opened 2026-05-05
[PB-6350]: feat/add monthly commitment plans logic

xabg2:
## What
We are introducing monthly plans with a 12-month commitment. Users can already purchase monthly plans, but we now need to handle cancellations according to this commitment period.

## How

To support this, we update the cancel_at property of the subscription to schedule cancellation based on the remaining months required to complete the 12-month commitment.
Additionally, a commitment object has been added to the subscription response. This includes:
- Whether the commitment is active
- The number of months remaining
- The scheduled cancellation date

### DIFF
--- a/src/services/payment.service.ts
+++ b/src/services/payment.service.ts
@@ -493,7 +493,7 @@ export class PaymentService {
     const cancelAtDate = createdAt.add(periodsElapsed + 1, 'year');
     const cancelAt = cancelAtDate.unix();
 
-    const isFirstMonth = monthsElapsed === 0 && now.diff(createdAt, 'day') < 30;
+    const isFirstMonth = monthsElapsed === 0 && now.diff(createdAt, 'day') <= 30;
     const remainingPayments = monthsIntoPeriod === 0 ? 12 : 12 - monthsIntoPeriod;
     const cancellationDate = cancelAtDate.toISOString();
 
--- a/src/services/payment.service.ts
+++ b/src/services/payment.service.ts
@@ -471,7 +471,46 @@ export class PaymentService {
     return { maxSpaceBytes: parseInt(price.metadata.maxSpaceBytes), recurring: isRecurring };
   }
 
+  /**
+   * Gets the annual commitment cancellation info
+   * @param subscription - The subscription we want to get info for
+   * @returns - The annual commitment cancellation info (remaining payments, amount per month, currency, cancel at)
+   */
+  getAnnualCommitmentCancellationInfo(subscription: Stripe.Subscription): {
+    remainingPayments: number;
+    cancelAt: number;
+    cancellationDate: string;
+  } {
+    const createdAt = new Date(subscription.created * 1000);
+    const now = new Date();
+
+    const monthsElapsed = (now.getFullYear() - createdAt.getFullYear()) * 12 + (now.getMonth() - createdAt.getMonth());
+    const monthsIntoPeriod = monthsElapsed % 12;
+    const periodsElapsed = Math.floor(monthsElapsed / 12);
+
+    const cancelAtDate = new Date(createdAt);
+    cancelAtDate.setFullYear(cancelAtDate.getFullYear() + periodsElapsed + 1);
+    const cancelAt = Math.floor(cancelAtDate.getTime() / 1000);
+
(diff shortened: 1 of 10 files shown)


tests/src/services/klaviyo.service.test.ts
```diff
@@ -37,11 +35,6 @@ describe('KlaviyoTrackingService', () => {
   });
 
   describe('Initialization', () => {
-    test('When instantiated without an API Key in config, then it throws an error', () => {
```
sg-gs, line 40: Don't remove the test, expect a log now

tests/src/services/payment.service.test.ts
```diff
@@ -1125,4 +1125,141 @@ describe('Payments Service tests', () => {
...
+
+    test('when the user just subscribed this month, then they have 12 remaining payments and cancel date is one year from now', () => {
+      const now = new Date('2026-05-05T10:00:00Z');
+      jest.setSystemTime(now);
+
+      const subscription = makeSubscriptionWithCommitment(new Date('2026-05-01T00:00:00Z'));
+      const { remainingPayments, cancelAt } = paymentService.getAnnualCommitmentCancellationInfo(subscription);
+
+      expect(remainingPayments).toBe(12);
+      expect(new Date(cancelAt * 1000).getFullYear()).toBe(2027);
```
sg-gs: I advise using `dayjs`, avoiding the manual date management, which is a pain

tests/src/services/payment.service.test.ts
```diff
@@ -1125,4 +1125,141 @@ describe('Payments Service tests', () => {
...
+        },
+      });
+
+      jest.spyOn(paymentService, 'getSubscriptionById').mockResolvedValue(subscription as any);
+      const updateSpy = jest.spyOn(stripe.subscriptions, 'update').mockResolvedValue(subscription as unknown as any);
+      const cancelSpy = jest.spyOn(stripe.subscriptions, 'cancel').mockResolvedValue(undefined as any);
+
+      await paymentService.cancelSubscription(subscription.id);
+
+      expect(updateSpy).toHaveBeenCalledWith(subscription.id, { cancel_at: expect.any(Number) });
```
sg-gs: Shouldn't you match the `cancel_at` date here to be the date you expect it to be? (end of the commitment period)

xabg2: we are checking that the subscription is updated with the cancel_at property and not cancelling the subscription directly. But yeah, we can be more specific and match the dates.

src/services/payment.service.ts
```diff
@@ -471,7 +472,47 @@ export class PaymentService {
...
+    const now = dayjs();
+
+    const monthsElapsed = now.diff(createdAt, 'month');
+    const monthsIntoPeriod = monthsElapsed % 12;
+    const periodsElapsed = Math.floor(monthsElapsed / 12);
+
+    const cancelAtDate = createdAt.add(periodsElapsed + 1, 'year');
+    const cancelAt = cancelAtDate.unix();
+
+    const isFirstMonth = monthsElapsed === 0 && now.diff(createdAt, 'day') < 30;
```
sg-gs: <= I think, isn't? It's up to 30 days.

xabg2: Right, changing it

tests/src/services/payment.service.test.ts
```diff
@@ -1125,4 +1127,283 @@ describe('Payments Service tests', () => {
...
+    test('when the user subscribed more than 30 days ago, then they are no longer in their first month', () => {
+      jest.setSystemTime(dayjs('2026-05-05T10:00:00Z').toDate());
+
+      const subscription = makeSubscriptionWithCommitment(dayjs('2026-03-01T10:00:00Z'));
+      const { isFirstMonth } = paymentService.getAnnualCommitmentCancellationInfo(subscription);
+
+      expect(isFirstMonth).toBe(false);
+    });
+
+    test('when the user has completed a full year and starts a new cycle, then they are not considered in their first month', () => {
```
sg-gs, line 1227: Nice corner case caught @xabg2

src/services/payment.service.ts
xabg2: Changed! @sg-gs

sg-gs (approved):

PR #366, opened 2026-05-05
[_]: refactor/b2b-cleanup

xabg2:
Removed the functions responsible for initializing and updating workspaces, as these plans are no longer offered.


src/controller/checkout.controller.ts
```diff
@@ -192,11 +190,14 @@ export function checkoutController(usersService: UsersService, paymentsService:
           throw new ForbiddenError();
         }
 
+        const price = await paymentsService.getPriceById(priceId);
+
+        if (price.type === UserType.Business) throw new BadRequestError('Business plan is not available');
```
sg-gs: Adding _no longer_ could make things clearer and easy to debug in case this error happens

src/services/payment.service.ts
```diff
@@ -140,20 +138,8 @@ export class PaymentService {
...
-    if (isBusinessProduct && minimumSeats && maximumSeats) {
-      if (seatsForBusinessSubscription > parseInt(maximumSeats)) {
-        throw new InvalidSeatNumberError('The new price does not allow the current amount of seats');
-      }
-
-      if (seatsForBusinessSubscription < parseInt(minimumSeats)) {
-        throw new InvalidSeatNumberError('The new price does not allow the current amount of seats');
-      }
-    }
+    if (isBusinessProduct) throw new BadRequestError('Business plan is not available');
```
sg-gs: Same here

src/services/payment.service.ts
```diff
@@ -167,7 +153,7 @@ export class PaymentService {
       items: [
         {
           price: priceId,
-          quantity: seats,
+          quantity: 1,
```
sg-gs, line 156: Why?

xabg2: Because now we only accept individual plans (1 product at once).

sg-gs (approved):

PR #367, opened 2026-05-11
[_]: refactor/prices

xabg2:
Refactored Stripe price handling by moving `getPrices` and `getPriceById` from `payments.service` to the Stripe provider/adapter. Updated related controllers and services to use the new implementation where needed.


xabg2:
This has been splitted in 3 other smaller PRs

PR #370, opened 2026-05-12
[_]: refactor/get price by id in checkout controller

xabg2:
Using the `getPriceById` and `getPrices` functions from the stripe adapter. I also removed the useless functions.

Things that I've checked:
- Create subscription
- Create payment intent
- Get a Price By It's ID
- Get the available prices


sg-gs (approved):
These changes involve field changes that if are wrong, could make delicate parts of the process fail (maxSpaceBytes, etc). Test them carefully again after deployment.

PR #375, opened 2026-06-22
[PB-5977]: feat/integrate MailService for downgrade and upgrade handling

jzunigax2:
Calls the new mail-server gateway suspend/reactivate endpoints so a user's mail account follows their plan state.

- **Reactivate** on `invoice.paid`: `InvoiceCompletedHandler` calls `applyMailFeatures` after Drive/VPN, gated on `featuresPerService[Service.Mail].enabled`.
- **Suspend** on plan cancellation/downgrade: `TiersService.removeTier` now handles `case Service.Mail` reached via `handleCancelPlan`.

---
⚠️ Introduces new env variables:
- `MAIL_URL`: base URL of the mail-server gateway
- `MAIL_GATEWAY_SECRET`: matching RSA key pair from mail's `GATEWAY_PUBLIC_SECRET` variable


src/services/objectStorage.service.ts
```diff
@@ -24,7 +17,7 @@ export class ObjectStorageService {
   }
 
   async reactivateAccount(payload: { customerId: string }): Promise<void> {
-    const jwt = signToken('5m', this.config.OBJECT_STORAGE_GATEWAY_SECRET);
+    const jwt = signGatewayToken('5m', this.config.OBJECT_STORAGE_GATEWAY_SECRET);
```
xabg2, line 20: 🚀

tests/src/services/mail.service.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const expectedParams = {
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: 'Bearer signed-jwt',
+    },
+  };
+
+  describe('suspendAccount', () => {
+    it('When called, then it POSTs to the suspend gateway endpoint with the signed token', async () => {
```
xabg2: Use test instead of it

tests/src/services/tiers.service.test.ts
```diff
@@ -369,4 +375,46 @@ describe('TiersService tests', () => {
       expect(removeVPNTierSpy).toHaveBeenCalledWith(uuid, tier.featuresPerService[Service.Vpn].featureId);
     });
   });
+
+  describe('Enable Mail access based on user tier', () => {
+    it('When Mail is enabled, then a request to reactivate the mail account is sent', async () => {
```
xabg2: Same

xabg2 (approved):

src/utils/signGatewayToken.ts
```diff
@@ -0,0 +1,10 @@
+import { sign } from 'jsonwebtoken';
+import { isProduction } from '../config';
+
+export function signGatewayToken(duration: string, secret: string): string {
+  return sign({}, Buffer.from(secret, 'base64').toString('utf8'), {
+    algorithm: 'RS256',
+    expiresIn: duration,
+    ...(!isProduction ? { allowInsecureKeySizes: true } : null),
```
sg-gs, line 8: Why is this needed?

src/services/mail.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import { signGatewayToken } from '../utils/signGatewayToken';
+
+export class MailService {
+  constructor(
+    private readonly config: AppConfig,
+    private readonly axios: AxiosInstance,
+  ) {}
+
+  async suspendAccount(uuid: User['uuid']): Promise<void> {
+    await this.axios.post(`${this.config.MAIL_URL}/gateway/accounts/${uuid}/suspend`, {}, this.requestConfig());
```
sg-gs, line 13: This is related to mail, not to here, but, shouldn't this be a PATCH as it is a partial update of the user enabled status field or so?

src/utils/signGatewayToken.ts
xabg2: BEcause there is a problem with signed tokens in development. It was causing problems in drive-server too some time ago, not sure if it was fixed in drive-server or still doing this

sg-gs: Mm, a bit weird, as 1 yr ago I tested the object storage - payments interconnection and therefore, signing locally and I did not have this issue

sg-gs:
Please @jzunigax2 on any PR that needs new environment variables: add a warning to the PR's description. Otherwise, I may ignore it and cause a downtime

src/services/mail.service.ts
xabg2: It also disables some mail features (e.g. disable send emails, update emails and destroy emails), not just updating a status field. Not sure if PATCH is better. POST is also accepted, don't you think?

src/utils/signGatewayToken.ts
xabg2: I added this fix like a year/2 years ago haha. He just extracted it to a function.
- [storage service](https://github.com/internxt/payments-server/blob/0fd166b64b0c7945e2596f68d66ea8d6ec6a3b6b/src/services/storage.service.ts#L11)
- [users service](https://github.com/internxt/payments-server/blob/0fd166b64b0c7945e2596f68d66ea8d6ec6a3b6b/src/services/users.service.ts#L20)

jzunigax2: yeah I guess we may have different env values. the ones I use for these keys were shared to me when I first joined and I never touched them since. This workaround also lives in drive server

src/services/mail.service.ts
jzunigax2: plus a post feels more appropiate to me as its an explicit action `/resource/id/suspend`, don't you think?

sg-gs:
Environment variables set @jzunigax2

sg-gs (approved):

PR #384, opened 2026-07-02
[PB-6370]: feat/apply cancellation trial

xabg2:
A new cancellation flow has been added for eligible users.

If a user has an active subscription that is still within the cancellation window (the first month), they can apply a 30-day trial to their subscription. This trial grants the user 30 additional free days and extends the cancellation window by the same period.

### DIFF
--- a/src/core/users/MongoDBUsersRepository.ts
+++ b/src/core/users/MongoDBUsersRepository.ts
@@ -2,10 +2,15 @@ import { Collection, MongoClient, ObjectId } from 'mongodb';
 import { User } from './User';
 
 import { UsersRepository } from './UsersRepository';
+import dayjs from 'dayjs';
 
 interface MongoUser extends Omit<User, 'customerId' | 'id'> {
   _id: ObjectId;
   customer_id: string;
+  cancellation_trial?: {
+    redeemed: boolean;
+    redeemed_at: Date;
+  };
 }
 
 export class MongoDBUsersRepository implements UsersRepository {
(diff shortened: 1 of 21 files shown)


src/services/payment.service.ts
```diff
@@ -435,20 +436,22 @@ export class PaymentService {
     const cancelAtDate = createdAt.add(periodsElapsed + 1, 'year');
     const cancelAt = cancelAtDate.unix();
 
-    const isFirstMonth = monthsElapsed === 0 && now.diff(createdAt, 'day') <= 30;
-    const remainingPayments = monthsIntoPeriod === 0 ? 12 : 12 - monthsIntoPeriod;
+    const isInFirstMonth = monthsElapsed === 0 && now.diff(createdAt, 'day') <= 30;
+
+    const isElegibleForCancellation = isInFirstMonth || (monthsElapsed === 1 && !!subscription.isTrialing);
```
sg-gs, line 440: What's the diff between `isInFirstMonth` and `monthsElapsed === 1` ?

xabg2: `isInFirstMonth` is, in fact, the first month (`monthsElapsed === 0`).
`monthsElapsed === 1` is the second month, so we check that the user is in the second month and has trial to allow him to cancel de sub as well.

sg-gs: Why months elapsed are in 0-based indexing?

src/core/users/MongoDBUsersRepository.ts
```diff
@@ -44,10 +45,43 @@ export class MongoDBUsersRepository implements UsersRepository {
     await this.collection.insertOne(this.userToMongoUser(user) as MongoUser);
   }
 
+  async redeemCancellationTrial(customerId: User['customerId']): Promise<boolean> {
```
sg-gs: These is the user's repository. You should use repository-related vocabulary, not business logic one. Instead of

`redeemCancellationTrial`

use 

`updateByCustomerId` or whatever applies here

src/core/users/MongoDBUsersRepository.ts
```diff
@@ -44,10 +45,43 @@ export class MongoDBUsersRepository implements UsersRepository {
...
+    const cancellationTrialKey: keyof UserDetails = 'cancellationTrial';
+    const cancellationTrial: UserDetails['cancellationTrial'] = {
+      redeemed: true,
+      redeemedAt: dayjs().toDate(),
+    };
+
+    return this.updateUserDetails(customerId, cancellationTrialKey, cancellationTrial);
+  }
+
+  async hasRedeemedCancellationTrial(customerId: User['customerId']): Promise<boolean> {
```
sg-gs: Use a findByCustomerId on the usecase and use the domain to check if the customer `hasRedeemedCancellationTrial`

src/services/payment.service.ts
xabg2: Because dayJS handles the months like this - it is called by doing `now.diff(createdAt, 'month')`. More info: https://day.js.org/docs/en/display/difference#list-of-all-available-units

sg-gs (approved):

PR #387, opened 2026-07-09
[PB-6371]: feature/expose endpoint for cancellation flow for users outside refund period

xabg2:
Exposing the endpoint to allow the users to cancel the subscription at the moment by paying the 50% of the remaining amount.

We also handle the logic in the `invoice.paid` webhook event to cancel the subscription.

### DIFF
--- a/src/services/payment.service.ts
+++ b/src/services/payment.service.ts
@@ -462,7 +462,7 @@ export class PaymentService {
       return;
     }
 
-    await stripePaymentsAdapter.deleteSubscription(subscriptionId);
+    await stripePaymentsAdapter.cancelSubscription(subscriptionId);
   }
 
   private calculateRemainingSubscriptionAmount(price: Price, remainingMonths: number): number {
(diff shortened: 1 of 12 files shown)


src/services/payment.service.ts
```diff
@@ -462,7 +462,7 @@ export class PaymentService {
       return;
     }
 
-    await this.provider.subscriptions.cancel(subscriptionId, {});
+    await stripePaymentsAdapter.deleteSubscription(subscriptionId);
```
sg-gs: I would keep the cancel wording, as the subscription status changes, the sub does not disappear

tests/src/controller/payments.controller.test.ts
```diff
@@ -628,4 +628,31 @@ describe('Payment controller e2e tests', () => {
       expect(body[0].interval).toBe('year');
     });
   });
+
+  describe('Cancel early charge', () => {
+    test('When the user wants to cancel the usb early, then we charge the 50% of the remaining amount', async () => {
```
sg-gs: Typo here, `usb` instead of `sub`

sg-gs (approved):

PR #390, opened 2026-07-20
[MKT_923]:feat/Relax required fields on POST /customer to match new checkout flow

jaaaaavier:
This PR Makes `customerName`, `lineAddress1`, `lineAddress2`, `city` and `postalCode` optional in the `POST /customer` body, keeping only `country` and `captchaToken` as required. The new checkout flow no longer collects full billing address up front for card payments, full address is only gathered for crypto payments. Matches the `CreateCustomerPayload` type already relaxed in `@internxt/sdk@1.17.19`.

### DIFF
--- a/src/controller/checkout.controller.ts
+++ b/src/controller/checkout.controller.ts
@@ -20,12 +20,12 @@ export function checkoutController(usersService: UsersService, paymentsService: 
 
     fastify.post<{
       Body: {
-        customerName: string;
-        lineAddress1: string;
-        lineAddress2: string;
-        city: string;
+        customerName?: string;
+        lineAddress1?: string;
+        lineAddress2?: string;
+        city?: string;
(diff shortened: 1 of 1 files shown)


sg-gs (approved):
LGTM although I do not know if this affects any B2B-related payment, as those params are usually required or nice-to-have in that case.

xabg2:
> LGTM although I do not know if this affects any B2B-related payment, as those params are usually required or nice-to-have in that case.

Afaik, b2b purchases are not supported anymore. At least I removed them from payments.  Only supported "update/remove".

xabg2 (approved):

jaaaaavier:
All I can tell you is that when I spoke to Fran, he told me that if any company needed this, it would be done manually. I think it’s likely that it will be updated in the future.

PR #391, opened 2026-07-23
[MKT_923]:fix/prevent checkout crash when Stripe cannot geolocate the user IP

jaaaaavier:
We were experiencing a problem where, when a user was geolocated in a region (for example Canada) that requires a sub-national jurisdiction (zip code and country) for tax calculations, we normally use the IP address, but in these cases that is insufficient. The error we encountered was: “The IP address provided is insufficient to determine the customer's tax location. Please provide an alternative address.”

To resolve this, we’ve made a couple of adjustments:
- Calculate taxes only when working with euros. This was discussed yesterday during a call and proposed because it appears to be too costly for us, it’s an additional measure beyond the fix.
- We updated the `calculateTax` function to allow it to return `undefined` if we encounter the specific error we’re discussing. To do this, we created the `isInsufficientLocationError` exception.
- Tests: Added two cases :
              1. IP cannot be geolocated → returns `undefined`
              2. another Stripe error → propagates.

### DIFF
--- a/src/controller/checkout.controller.ts
+++ b/src/controller/checkout.controller.ts
@@ -376,7 +376,7 @@ export function checkoutController(usersService: UsersService, paymentsService: 
           }
         }
 
-        if (userAddress || (postalCode && country) || user?.customerId) {
+        if (currency === 'eur' && (userAddress || (postalCode && country) || user?.customerId)) {
           taxForPrice = await paymentsService.calculateTax(
             priceId,
             amount,
@@ -389,7 +389,7 @@ export function checkoutController(usersService: UsersService, paymentsService: 
         }
 
         const taxAmount = taxForPrice?.tax_amount_exclusive ?? 0;
-        const amountTotal = taxForPrice?.amount_total ?? price.amount;
+        const amountTotal = taxForPrice?.amount_total ?? amount;
 
         return res.status(200).send({
           price: price.toJSON(),
--- a/src/controller/checkout.controller.ts
+++ b/src/controller/checkout.controller.ts
@@ -363,6 +363,9 @@ export function checkoutController(usersService: UsersService, paymentsService: 
 
         const price = await stripePaymentsAdapter.getPriceById(priceId, currency);
 
+        const isEuroCurrency = currency === 'eur';
+        const isUserAddressProvided = !!userAddress || (!!postalCode && !!country) || !!user?.customerId;
+
         let amount = price.amount;
 
         if (promoCodeName) {
@@ -376,7 +379,7 @@ export function checkoutController(usersService: UsersService, paymentsService: 
           }
         }
 
-        if (currency === 'eur' && (userAddress || (postalCode && country) || user?.customerId)) {
+        if (isEuroCurrency && isUserAddressProvided) {
           taxForPrice = await paymentsService.calculateTax(
             priceId,
             amount,
--- a/src/services/payment.service.ts
+++ b/src/services/payment.service.ts
@@ -45,6 +45,7 @@ import {
   Subscription as SubscriptionEntity,
 } from '../infrastructure/domain/entities/subscription';
 import { SUBSCRIPTION_EARLY_CANCELLATION_KEY } from '../constants';
+import { STRIPE_TAX_LOCATION_INVALID_CODE } from '../errors/stripeErrorCodes';
 import { Price } from '../infrastructure/domain/entities/price';
 
 export class PaymentService {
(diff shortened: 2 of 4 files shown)


src/services/payment.service.ts
```diff
@@ -1039,19 +1040,34 @@ export class PaymentService {
...
+          },
+        ],
+        currency,
+      });
+    } catch (error) {
+      const isInsufficientLocationError =
+        usingIpAddress &&
+        !customerId &&
+        error instanceof Stripe.errors.StripeError &&
+        error.message?.includes('insufficient to determine');
```
xabg2: Is there another way to check that the error is about the invalid IP? matching it using includes is not a good idea. Maybe it has a Code error or smth. Maybe `customer_tax_location_invalid` can help you.

tests/src/services/payment.service.test.ts
```diff
@@ -725,6 +725,32 @@ describe('Payments Service tests', () => {
 
       expect(taxes).toStrictEqual(mockedTaxes);
     });
+
+    it('When the IP cannot be geolocated to a tax jurisdiction, then no tax is returned so the price still loads', async () => {
```
xabg2: Use test, please

src/services/payment.service.ts
```diff
@@ -1039,19 +1040,34 @@ export class PaymentService {
...
+            tax_behavior: 'exclusive',
+            quantity: 1,
+          },
+        ],
+        currency,
+      });
+    } catch (error) {
+      const isInsufficientLocationError =
+        usingIpAddress &&
+        !customerId &&
```
xabg2: Why?

src/controller/checkout.controller.ts
```diff
@@ -376,7 +376,7 @@ export function checkoutController(usersService: UsersService, paymentsService:
           }
         }
 
-        if (userAddress || (postalCode && country) || user?.customerId) {
+        if (currency === 'eur' && (userAddress || (postalCode && country) || user?.customerId)) {
```
xabg2: Extract this if to a constant. Both the currency check and the other part.

src/services/payment.service.ts
```diff
@@ -1026,8 +1026,9 @@ export class PaymentService {
     customerId?: string,
     postalCode?: string,
     country?: string,
-  ): Promise<Stripe.Tax.Calculation> {
+  ): Promise<Stripe.Tax.Calculation | undefined> {
     const customerDetails: Stripe.Tax.CalculationCreateParams.CustomerDetails = {};
+    const usingIpAddress = !(postalCode && country);
```
xabg2: Use conventional names for booleans. E.g.:
- `isUsingIpAddress`
- `shouldUseIpAddress`

src/services/payment.service.ts
```diff
@@ -1039,19 +1040,34 @@ export class PaymentService {
...
+      });
+    } catch (error) {
+      const isInsufficientLocationError =
+        usingIpAddress &&
+        !customerId &&
+        error instanceof Stripe.errors.StripeError &&
+        error.message?.includes('insufficient to determine');
+
+      if (isInsufficientLocationError) {
+        Logger.warn(`Tax calculation skipped: IP ${ipAddress} is insufficient to determine tax location`);
```
xabg2: If I'm not mistaken, warn logs don't appear in Graylog. Use either `info` or `error`.

xabg2 (changes requested):

src/services/payment.service.ts
jaaaaavier: Okay, `info` better than `error`??

src/services/payment.service.ts
jaaaaavier: Okay👌

src/services/payment.service.ts
xabg2: As you are throwing and printing an error, maybe `error` is better.

src/controller/checkout.controller.ts
jaaaaavier: Okay👌

src/services/payment.service.ts
jaaaaavier: Perfect

src/services/payment.service.ts
jaaaaavier: Yes better

sg-gs (dismissed):
Leaving this up to @xabg2. The changes requested by him are almost the same things I could mention in this case. Special mention to the constants, as some ifs are ununderstandable right now.

src/services/payment.service.ts
jaaaaavier: Those checks have been removed. The workaround should be applied whenever Stripe cannot determine the tax jurisdiction, regardless of whether the information comes from an IP address or a stored customer address. I thought that perhaps if the customer had their address saved because they were a long-time user, but the truth is that the same error could still occur. The check is now simply:


`error.code === STRIPE_TAX_LOCATION_INVALID_CODE.`

xabg2 (approved):

PR #394, opened 2026-07-28
[MKT_949]:feat/cloudflare turnstile integration

jaaaaavier:
### Changes
The purpose of this PR is to integrate Cloudflare Turnstile into the repo. To do this, we’ve done the following:
- Create the `assertCaptcha` file. Here, we basically decide whether to use the new `verifyTurnstile` function or the old `verifyRecaptcha` function. This allows us to keep what we already had and know works as a fallback
- In the `verifyTurnstile` file, we’ve set up everything needed to verify with Cloudflare that the token is correct. This file was created based on the official Cloudflare Turnstile documentation.
- I’ve also extracted the `isTransientNetworkError` method to the `networkRetry` file, since we’ll be using it in different parts of the code.
- In the checkout controller, we’ve replaced `verifyRecaptcha` with the `asserCaptcha` described above, marked `captchaToken` as optional, and added `turnstileToken` in the same way; now, one of these two must be required
- New variables have been added
- Tests have been updated

### Test

Cloudflare publishes a set of **dummy keys**: public sitekey/secret pairs that force a deterministic
outcome (always pass, always fail, force an interactive challenge) while still going through the real
`siteverify` API. They let us exercise the whole chain — browser widget → drive-web → payments →
Cloudflare — without creating a widget in the dashboard.

Each scenario was run through a full checkout, checking three places: the request payload in the
browser, the payments logs, and the resulting HTTP status.

| Scenario | Keys | Result |
| --- | --- | --- |
| Happy path | sitekey `1x…BB` (pass, invisible) + secret `1x…AA` (pass) | `POST /checkout/customer` and `POST /checkout/payment-intent` both carry a `turnstileToken` alongside the reCAPTCHA `captchaToken`. No fallback warning in the logs. Payment completes. |
| Turnstile rejects the token | secret `2x…AA` (always fails) | `Turnstile rejected the token: invalid-input-response` → 403, and reCAPTCHA is never consulted. `responseTime: 60ms` confirms no retries were attempted, i.e. the code was correctly treated as terminal rather than transient. |
| Widget fails in the browser | sitekey `2x…BB` (always fails, invisible) | Console shows Turnstile error `600010`; no `turnstileToken` in the payload; payment completes through reCAPTCHA. |
| Misconfigured sitekey | invalid sitekey | Widget never initialises, `generateTurnstileToken()` returns `undefined`, checkout is unaffected and falls back to reCAPTCHA. |

I can add some images or video in the task if needed but i think we will need some QA

### Notes
The enviroment variables required are:

`TURNSTILE_ENDPOINT=https://challenges.cloudflare.com/turnstile/v0/siteverify`
`TURNSTILE_SECRET=the one used on ai-server`


sg-gs:
The markdown table format seems to not be correct @jaaaaavier

sg-gs (dismissed):
Please add the list of new environment variables in the PR's description @jaaaaavier

src/controller/checkout.controller.ts
xabg2, line 1: if `turnstileToken` is now the first option, maybe it should be required, not optional. WDYT?

src/utils/verifyTurnstile.ts
```diff
@@ -0,0 +1,70 @@
+import axios from 'axios';
+import { encode } from 'node:querystring';
+import config from '../config';
+import Logger from '../Logger';
+import { isTransientNetworkError, sleep } from './networkRetry';
+
+const MAX_TURNSTILE_ATTEMPTS = 3;
+const TURNSTILE_RETRY_BASE_DELAY_MS = 300;
+const RETRYABLE_TURNSTILE_ERROR_CODE = 'internal-error';
```
xabg2: Maybe we can also extract the shared constants between this file and the captcha one to a `constants.ts`.

src/utils/verifyTurnstile.ts
```diff
@@ -0,0 +1,70 @@
...
+
+type TurnstileVerificationResponse = {
+  success: boolean;
+  'error-codes'?: string[];
+  challenge_ts?: string;
+  hostname?: string;
+};
+
+export class TurnstileUnavailableError extends Error {}
```
xabg2: Add the necessary info here with:
```
super(message = '...') {
this.message = message;
this.name = 'TurnstileUnavailableError';
Object.setPrototypeOf(this, TurnstileUnavailableError.prototype);
}
```

Context: https://github.com/internxt/payments-server/blob/master/src/errors/UsersErrors.ts#L12

tests/src/utils/assertCaptcha.test.ts
```diff
@@ -0,0 +1,95 @@
...
+});
+
+jest.mock('../../../src/Logger', () => ({
+  __esModule: true,
+  default: {
+    warn: jest.fn(),
+    info: jest.fn(),
+    error: jest.fn(),
+  },
+}));
```
xabg2: I think you can use `spyOn` for Logger instead of mocking it.

src/utils/verifyTurnstile.ts
```diff
@@ -0,0 +1,70 @@
...
+  return res.data.success;
+}
+
+export async function verifyTurnstile(turnstileToken: string): Promise<boolean> {
+  for (let attempt = 0; attempt < MAX_TURNSTILE_ATTEMPTS; attempt++) {
+    try {
+      return await requestTurnstile(turnstileToken);
+    } catch (err) {
+      const isLastAttempt = attempt === MAX_TURNSTILE_ATTEMPTS - 1;
+      const isRetryable = isTransientNetworkError(err) || err instanceof TurnstileUnavailableError;
```
xabg2: Extract this to a function. Something like:
```
function isRetryableTurnstileError(err: unknown): boolean {
  return isTransientNetworkError(err) || err instanceof TurnstileUnavailableError;
}
```

xabg2 (approved):

sg-gs (approved):
Adding the environment variables @jaaaaavier

sg-gs:
Secrets updated @jaaaaavier 

Turnstile Widget updated to consider `drive.internxt.com` also.

PR #395, opened 2026-07-28
[_]: fix/use startDate instead of created

xabg2:
We're changing the parameter we use to determine when a subscription began:
- `created` (previously used): the date when the subscription object was created in Stripe.
- `startDate` (used now): the date when the subscription actually started.

With this change, we can properly test the new cancellation flow. For example, we can manually create a subscription in the Stripe Dashboard with a start date in the past, allowing the commitment period to be calculated correctly.


sg-gs (approved):
Nice

PR #396, opened 2026-07-28
[_]: fix/get user payment method

xabg2:
Retrieve a list of payment methods giving the user. It will only be used as a fallback.


xabg2:
No needed for now.

PR #397, opened 2026-08-04
fix(mail): swallow 404 errors, reintroduce commented out mail related methods

jzunigax2:


src/services/mail.service.ts
```diff
@@ -10,11 +11,24 @@ export class MailService {
...
   async reactivateAccount(uuid: User['uuid']): Promise<void> {
-    await this.axios.post(`${this.config.MAIL_URL}/gateway/accounts/${uuid}/reactivate`, {}, this.requestConfig());
+    await this.post(`/gateway/accounts/${uuid}/reactivate`, uuid);
+  }
+
+  private async post(path: string, uuid: User['uuid']): Promise<void> {
+    try {
+      await this.axios.post(`${this.config.MAIL_URL}${path}`, {}, this.requestConfig());
+    } catch (error) {
+      if (isAxiosError(error) && error.response?.status === 404) {
```
xabg2, line 25: If we get a 404 from Mail for this EP, can it only happen because the mail user doesn’t exist? Just want to make sure. If not, maybe we could use the specific error code, e.g. `UserNotFoundError` or whatever is appropriate.

jzunigax2: yes that is the only case where those endpoints would return a 404

xabg2 (approved):

